PR #12, opened 2023-08-04
Generalized Ray `put`/`get`

omus:
Allows us to store any bytes in the object store.

---
  - https://app.asana.com/0/0/1205184395990358

### DIFF
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -40,7 +40,7 @@ ObjectID put(void *ptr, size_t size) {
 }
 
 // https://github.com/ray-project/ray/blob/a4a8389a3053b9ef0e8409a55e2fae618bfca2be/src/ray/core_worker/test/core_worker_test.cc#L210-L220
-void *get(ObjectID object_id) {
+std::shared_ptr<Buffer> get(ObjectID object_id) {
     auto &driver = CoreWorkerProcess::GetCoreWorker();
 
     // Retrieve our data from the object store
@@ -53,16 +53,12 @@ void *get(ObjectID object_id) {
         return nullptr;
     }
 
-    return result->GetData()->Data();
+    return result->GetData();
 }
 
 std::string ToString(ray::FunctionDescriptor function_descriptor)
 {
     return function_descriptor->ToString();
-}
-
-LocalMemoryBuffer *demo(LocalMemoryBuffer *buffer) {
-    return buffer;
 }
 
 namespace jlcxx
@@ -73,11 +69,15 @@ namespace jlcxx
 
 JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
 {
+    // WARNING: The order in which register types and methods with jlcxx is important.
+    // You must register all function arguments and return types with jlcxx prior to registering
+    // the function. If you fail to do this you'll get a "No appropriate factory for type" upon
+    // attempting to use the shared library in Julia.
+
     mod.method("initialize_coreworker", &initialize_coreworker);
     mod.method("shutdown_coreworker", &shutdown_coreworker);
     mod.add_type<ObjectID>("ObjectID");
     mod.method("put", &put);
-    mod.method("get", &get);
 
     // enum Language
     mod.add_bits<ray::Language>("Language", jlcxx::julia_type("CppEnum"));
@@ -117,8 +117,7 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
     mod.add_type<LocalMemoryBuffer>("LocalMemoryBuffer", jlcxx::julia_base_type<Buffer>())
         .constructor<uint8_t *, size_t, bool>();
 
-    // mod.method("demo", &demo);
-    mod.method("demo", [] (LocalMemoryBuffer *b) { return demo(b); });
+    mod.method("get", &get);
 
     // mod.add_type<RayObject>("RayObject")
     //     .constructor<const std::shared_ptr<Buffer>,
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -65,6 +65,11 @@ std::shared_ptr<Buffer> shared_buffer(Buffer *buffer)
     return std::shared_ptr<Buffer>(buffer);
 }
 
+std::shared_ptr<LocalMemoryBuffer> make_shared_local_memory_buffer(uint8_t *data, size_t size, bool copy_data = false)
+{
+    return std::make_shared<LocalMemoryBuffer>(data, size, copy_data);
+}
+
 namespace jlcxx
 {
     // Needed for upcasting
@@ -147,6 +152,8 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
         // .constructor<uint8_t *, size_t, bool>([] (uint8_t *data, size_t size, bool copy_data = false) {
         //     return jlcxx::create<LocalMemoryBuffer>(data, size, copy_data);
         // });
+
+    mod.method("make_shared_local_memory_buffer", &make_shared_local_memory_buffer);
 
     /*
     mod.add_type<rpc::ObjectReference>("ObjectReference")
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -126,9 +126,13 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
     // https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/common/buffer.h
     mod.add_type<Buffer>("Buffer")
         .method("_data_pointer", &Buffer::Data)
-        .method("_sizeof", &Buffer::Size)  // TODO: How can we extend a method in Base?
+        .method("_sizeof", &Buffer::Size)
         .method("owns_data", &Buffer::OwnsData)
         .method("is_plasma_buffer", &Buffer::IsPlasmaBuffer);
+        // .method("_data_pointer", [] (std::shared_ptr<Buffer> b) { b->Data(); })
+        // .method("owns_data", [] (std::shared_ptr<Buffer> b) { b->OwnsData(); })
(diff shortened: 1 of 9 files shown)


omus:
Tracked the issue down to a double deallocation. CxxWrap doesn't like it when you allocate an object and then use `std::shared_ptr` later as CxxWrap adds finalizers for both of these Julia objects resulting in a double deallocation. Some useful code for figuring this out:
```julia
using CxxWrap
using ray_core_worker_julia_jll: ray_core_worker_julia_jll, initialize_coreworker, shutdown_coreworker
using ray_core_worker_julia_jll: LocalMemoryBuffer, data_pointer, get, put, shared_ptr, ObjectID, make_shared_local_memory_buffer

data = UInt16[1:3;]
lmb = LocalMemoryBuffer(Ptr{Nothing}(pointer(data)), sizeof(data), true)
lmb = nothing
Base.GC.gc()  # Works

lmb = LocalMemoryBuffer(Ptr{Nothing}(pointer(data)), sizeof(data), true)
sptr = shared_ptr(CxxPtr(lmb))
sptr.cpp_object
sptr = nothing
Base.GC.gc()  # Works

slmb = make_shared_local_memory_buffer(Ptr{Nothing}(pointer(data)), sizeof(data), true)
slmb.cpp_object
slmb = nothing
Base.GC.gc()  # Works

lmb = LocalMemoryBuffer(Ptr{Nothing}(pointer(data)), sizeof(data), true)
sptr = shared_ptr(CxxPtr(lmb))
sptr.cpp_object
sptr = nothing
lmb = nothing
Base.GC.gc()  # Deallocates for the `shared_ptr` and the `LocalMemoryBuffer*`
```
Requires the use of a custom finalizer to determine what is being deallocated.

omus:
Recap on debugging the segfault issue:

By using a [custom finalizer](https://github.com/JuliaInterop/CxxWrap.jl/tree/main#overriding-finalization-behavior) and a MWE it became easy to determine the core issue was that we were deallocating the same C++ instance twice. The reason for this was that we created a `LocalMemoryBuffer` from Julia which was allocated and then turned it into a `std::shared_ptr`. When both of these were garbage collected we ended up trying to release already released memory.

The solution to this was to modify the `LocalMemoryBuffer` constructor to always return the `std::shared_ptr` (taking inspiration from: https://github.com/JuliaInterop/CxxWrap.jl/issues/141#issuecomment-491373720). It would be nice to be able to have constructors for a non-smart pointer and the smart-pointer but for now having just the smart pointer is more convenient for working with the shared object store.

glennmoy (approved):

deps/wrapper.cc
```diff
@@ -104,6 +121,26 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
...
+    });
+
+    mod.method("put", &put);
+    mod.method("get", &get);
+
+    // mod.add_type<RayObject>("RayObject")
+    //     .constructor<const std::shared_ptr<Buffer>,
+    //                  const std::shared_ptr<Buffer>,
+    //                  const std::vector<rpc::ObjectReference>,
+    //                  bool>();
```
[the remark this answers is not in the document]
omus: We'll probably want to expose this in the future. It's not technically needed here but I started down this path so there's little harm to leaving it in in case we do end up using it

deps/wrapper.cc
```diff
@@ -57,25 +49,50 @@ std::string get(ObjectID object_id) {
...
+    template<typename T>
+    struct Finalizer<T, SpecializedFinalizer>
+    {
+        static void finalize(T* to_delete)
+        {
+            std::cout << "calling delete on: " << to_delete << std::endl;
+            delete to_delete;
+        }
+    };
+}
```
kleinschmidt, line 84: is htis still necessary?  IIUC it was used to debug the segfaults but those are taken care of now...

deps/wrapper.cc
```diff
@@ -1,12 +1,5 @@
 #include "wrapper.h"
 
-namespace julia {
-
-using namespace ray;
-using ray::core::CoreWorkerProcess;
-using ray::core::CoreWorkerOptions;
-using ray::core::WorkerType;
-
```
kleinschmidt, line 9: these were moved to the header file right?  except for teh namespace.  why remove the namespace?

deps/wrapper.cc
```diff
@@ -57,25 +49,50 @@ std::string get(ObjectID object_id) {
...
+        }
+    };
+}
+
 JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
 {
+    // WARNING: The order in which register types and methods with jlcxx is important.
+    // You must register all function arguments and return types with jlcxx prior to registering
+    // the function. If you fail to do this you'll get a "No appropriate factory for type" upon
+    // attempting to use the shared library in Julia.
```
kleinschmidt, line 91: nice

deps/wrapper.cc
```diff
@@ -104,6 +121,26 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
...
+    // https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/common/buffer.h
+    mod.add_type<Buffer>("Buffer")
+        .method("Data", &Buffer::Data)
+        .method("Size", &Buffer::Size)
+        .method("OwnsData", &Buffer::OwnsData)
+        .method("IsPlasmaBuffer", &Buffer::IsPlasmaBuffer);
+    mod.add_type<LocalMemoryBuffer>("LocalMemoryBuffer", jlcxx::julia_base_type<Buffer>());
+    mod.method("LocalMemoryBuffer", [] (uint8_t *data, size_t size, bool copy_data = false) {
+        return std::make_shared<LocalMemoryBuffer>(data, size, copy_data);
+    });
```
kleinschmidt, line 135: just to make sure I'm understanding what this is doing:

first line wraps the ray LocalMemoryBuffer type define dhere: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/common/buffer.h#L56

second adds a julia method call `LocalMemoryBuffer` that returns a shared pointer to an instance of the local memory buffer, forwarding the data/size/copy arguments to the _real_ constructor (which we're not exposing).  is that right?

```suggestion
    // class LocalMemoryBuffer
    // https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/common/buffer.h#L56
    mod.add_type<LocalMemoryBuffer>("LocalMemoryBuffer", jlcxx::julia_base_type<Buffer>());
    mod.method("LocalMemoryBuffer", [] (uint8_t *data, size_t size, bool copy_data = false) {
        return std::make_shared<LocalMemoryBuffer>(data, size, copy_data);
    });
```

test/buffer.jl
```diff
@@ -0,0 +1,70 @@
...
+    data = UInt16[1:3;]
+
+    @testset "non-copied object reference" begin
+        buffer = LocalMemoryBuffer(pointer_from_objref(data), sizeof(data), false)
+        @test Data(buffer[]) == pointer_from_objref(data)
+        @test Size(buffer[]) == sizeof(data)
+        @test !OwnsData(buffer[])
+        @test !IsPlasmaBuffer(buffer[])
+
+        buffer_ptr = Data(buffer[]).cpp_object
```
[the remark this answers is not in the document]
kleinschmidt: > If we added a `unsafe_pointer_to_objref(::CxxPtr)`

I think this would be useful to do, maybe as an upstream PR to CxxWrap?  referring to the wrapped pointer seems dangerous.

test/buffer.jl
```diff
@@ -0,0 +1,70 @@
...
+using ray_core_worker_julia_jll: LocalMemoryBuffer, Data, Size, OwnsData, IsPlasmaBuffer
+
+@testset "LocalMemoryBuffer" begin
+    data = UInt16[1:3;]
+
+    @testset "non-copied object reference" begin
+        buffer = LocalMemoryBuffer(pointer_from_objref(data), sizeof(data), false)
+        @test Data(buffer[]) == pointer_from_objref(data)
+        @test Size(buffer[]) == sizeof(data)
+        @test !OwnsData(buffer[])
```
kleinschmidt, line 12: what are the semantics of this?  set on construction (if teh constructor copies the data into its own memory)?

ah yeah, here we go: https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/common/buffer.h#L69-L70

test/buffer.jl
```diff
@@ -0,0 +1,70 @@
...
+        buffer = LocalMemoryBuffer(pointer_from_objref(data), sizeof(data), true)
+        @test Data(buffer[]) != pointer_from_objref(data)
+        @test Size(buffer[]) == sizeof(data)
+        @test OwnsData(buffer[])
+        @test !IsPlasmaBuffer(buffer[])
+
+        # Attempting to use:
+        # `buffer_ptr = Data(buffer[]).cpp_object; unsafe_pointer_to_objref(buffer_ptr)`
+        # would result in a segfault.
+    end
```
kleinschmidt, line 34: I'm vaguely concerned about this...seems like at some point we'll need to be able to construct these buffers from julia objects right?  or should we be serializing them _first_ and then creating the buffer from the array?

test/buffer.jl
```diff
@@ -0,0 +1,70 @@
...
+        @test Size(buffer[]) == sizeof(data)
+        @test OwnsData(buffer[])
+        @test !IsPlasmaBuffer(buffer[])
+
+        buffer_ptr = Data(buffer[]).cpp_object
+        buffer_size = Size(buffer[])
+        T = eltype(data)
+        len = Size(buffer[]) ÷ sizeof(T)
+        result = Vector{T}(undef, len)
+        unsafe_copyto!(Ptr{UInt8}(pointer(result)), buffer_ptr, buffer_size)
```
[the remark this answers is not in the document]
kleinschmidt: wait, there are 40 extra bytes for the array uhh header stuff before teh actual contents??

PR #15, opened 2023-08-08
Build Ray CLI from source in CI

omus:
Follow up to #8


omus:
Possibly this is due to not installing the dashboard?

omus:
Cache seemed to be ignored on the last commit. Will need to re-run the jobs after they complete to see if the `npm build` addition helped at all

omus:
It appears that the `copy_to_workspace` rules aren't the only copied data that is required:
```
error: [Errno 2] No such file or directory: 'ray/core/src/ray/gcs/gcs_server'
``` 
I'll switch over to using a `git ls-files --others --ignored --exclude-standard` approach to attempt and capture these files.

omus:
Restoring the cache from an alternate Julia version doesn't seem to speed up the build at all. I'll remove that possibility from the restore-key list (https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5829330794/job/15808647370?pr=15)

.github/workflows/CI.yml
```diff
@@ -45,29 +47,57 @@ jobs:
...
+
+          # The Ray BUILD.bazel includes a bunch of `copy_to_workspace` rules which copy build output
+          # into the Ray worktree. When we only restore the Bazel cache then re-building causes these
+          # rules to be skipped resulting in `error: [Errno 2] No such file or directory`. By manually
+          # saving/restoring these files we can work around this.
+          RAY_GEN_CACHE_DIR=~/.cache/ray-generated
+          if [ -d "$RAY_GEN_CACHE_DIR" ]; then
+              dest=$(pwd)
+              pushd $RAY_GEN_CACHE_DIR
+              cp -rp --parents \
```
glennmoy, line 75: `--parents` only seems to be available in linux - I guess this precludes us testing the build on macos?

.github/workflows/CI.yml
```diff
@@ -45,29 +47,57 @@ jobs:
...
-        run: pip install -U "ray[default]==2.5.1" "pydantic<2"
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
 
       # https://github.com/actions/cache/tree/main/save#always-save-cache
-      - uses: actions/cache/save@v3
-        if: always()
+      - name: Save build cache
+        uses: actions/cache/save@v3
+        if: always() && steps.build-cache.outputs.cache-hit != 'true'
```
glennmoy, line 100: what does the second condition check? something about the `build-cache` step not succeeding?

glennmoy (approved):

.github/workflows/CI.yml
omus: We can update this in the future to use multiple steps. We would just need to make the directory first and then perform the copies. Using `--parents` should be faster as we can avoid having to perform multiple copy commands. Most likely we'd just have an alternate code path for macOS

omus:
I did some experiments to see how well the CI runs scale with larger runners. Using the [GitHub-hosted pricing information](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) we can determine how much each run costs:

- GitHub standard runner, no cache: [1h16m30s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5829330794/job/15808647370)
- GitHub standard runner, cached: [7m0s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5829330794/job/15808647495)
- GitHub-hosted 8 core, no cache: [25m58s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5833804085) @ $0.032 = $0.832
- GitHub-hosted 8 core, cached: [2m10s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5833804085/job/15822852254) @ $0.032 = $0.096
- GitHub-hosted 16 core, no cache: [15m55s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5832552333/job/15818204565?pr=15) @ $0.064 = $1.024
- GitHub-hosted 16 core, cached: [1m50s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5832552333/job/15818685991) @ $0.064 = $0.128
- GitHub-hosted 32 core, no cache: [11m19s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5833497966/job/15821078772) @ $0.128 = $1.536
- GitHub-hosted 32 core, cached: [2m1s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5833497966/job/15821455052) @ $0.128 = $0.384
- GitHub-hosted 64 core, no cache: [9m20](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5833658093/job/15821553492) @ $0.256 = $2.56
- GitHub-hosted 64 core, cached: [2m9s](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5833658093/job/15821873425) @ $0.256 = $0.768

The larger runners help for cache save/restore, Bazel builds, and Julia package installs but don't provide us any current benefits for actually running the tests. It definitely seems like 16 core is the sweet spot so we'll stick with that.

PR #16, opened 2023-08-08
expose a `GcsClient` wrapper to julia

kleinschmidt:
Function manager will put/get stuff into the GCS internal key-value store, so a first step is being able to expose the KV store interface to julia.  Unfortunately this is not SO straightforward: the basic `GcsClient` needs a `io_service` (actually a `instrumented_io_service` but same thing basically) to manage the actual async execution of the requests, and the client that's embedded in the `CoreWorkerProcess` itself is private and not accessible via member functions (at least at the level we need) so we can't piggyback on that.  

My solution here is to add a `JuliaGcsClient` wrapper that manages the IO service and the associated thread (based on the GCS client test harness code).  ~~As you can see from my `Disconnect` function, I'm struggling to figure out how to tear this all down when it hasn't been connected yet (it segfaults even checking whether the ptr for the IO service is null or not).~~  EDIT: sorted.

Another possible direction would be to use the `PythonGcsClient` which AFAICT would work just fine for our purposes but is marked for "only Python/Cython use", whatever that means.

But the basic get/put functionality works just fine, and I think this is probably okay for now for the purposes of setting up the function manager stuff.

### DIFF
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -61,30 +61,30 @@ std::string ToString(ray::FunctionDescriptor function_descriptor)
 }
 
 JuliaGcsClient::JuliaGcsClient(const gcs::GcsClientOptions &options)
-  : options_(options) {
+    : options_(options) {
 }
 
 JuliaGcsClient::JuliaGcsClient(const std::string &gcs_address) {
-  options_ = gcs::GcsClientOptions(gcs_address);
+    options_ = gcs::GcsClientOptions(gcs_address);
 }
 
 Status JuliaGcsClient::Connect() {
-  gcs_client_ = std::make_unique<gcs::PythonGcsClient>(options_);
-  return gcs_client_->Connect();
+    gcs_client_ = std::make_unique<gcs::PythonGcsClient>(options_);
+    return gcs_client_->Connect();
 }
 
 std::string JuliaGcsClient::Get(const std::string &ns,
                                 const std::string &key,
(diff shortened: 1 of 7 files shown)


glennmoy (approved):

deps/wrapper.cc
```diff
@@ -69,6 +69,61 @@ std::string ToString(ray::FunctionDescriptor function_descriptor)
...
+    io_service_->stop();
+  }
+  if (io_service_thread_ && (io_service_thread_->joinable())) {
+    io_service_thread_->join();
+  }
+  if (gcs_client_) {
+    gcs_client_->Disconnect();
+    gcs_client_.reset();
+  }
+}
```
omus, line 102: I've been using 4-space indentations in our C code. Do we want to make a choice on that now?

omus (commented):
Would like to see a basic test added just to validate this works as intended.

deps/wrapper.cc
kleinschmidt: 4 is fine, I just need to configure my editor

kleinschmidt:
okay, after reviewing the ray code a bit more (and starting on some tests) I've deicded that wrapping the `PythonGcsClient` is probably a better path forward for us; it's a much simpler wrapper around the GRPC channel/requests/stubs and does not mess w/ the work thread etc. stuff.  I'll rework this to reflect that...

omus (approved):
Loving those tests

PR #18, opened 2023-08-10
set up local Ray.jl

kleinschmidt:
We have a repo for the user-facing Ray.jl already (https://github.com/beacon-biosignals/Ray.jl) but having separate repos for these two tightly coupled julia packages is inconvenient for initial rapid iteration.

Not _totally_ sure why we don't get QUITE as nice cache behavior as with the basic jll but it does go faster than the full build.


.github/workflows/Ray_CI.yml
```diff
@@ -0,0 +1,93 @@
+name: Ray.jl - CI
```
omus, line 1: Can we rename the other CI workflow to include the package name just so these match?

.github/workflows/Ray_CI.yml
```diff
@@ -0,0 +1,93 @@
...
+      # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
+      - uses: actions/cache/restore@v3
+        id: bazel-cache
+        with:
+          key: bazel-cache.${{ matrix.os }}.${{ matrix.arch }}.julia-${{ matrix.version }}.ray-${{ steps.ray-commit.outputs.head_sha }}.hash-${{ hashFiles('deps/WORKSPACE.bazel.tpl', 'deps/BUILD.bazel') }}
+          path: ~/.cache/bazel
+          restore-keys: |
+            bazel-cache.${{ matrix.os }}.${{ matrix.arch }}.julia-${{ matrix.version }}.ray-${{ steps.ray-commit.outputs.head_sha }}.
+            bazel-cache.${{ matrix.os }}.${{ matrix.arch }}.julia-${{ matrix.version }}.
+            bazel-cache.${{ matrix.os }}.${{ matrix.arch }}.
```
omus, line 71: The GHA cache action doesn't like to share caches between concurrently running jobs I would suggest doing:
```suggestion
          key: bazel-cache.ray-jl.${{ matrix.os }}.${{ matrix.arch }}.julia-${{ matrix.version }}.ray-${{ steps.ray-commit.outputs.head_sha }}.hash-${{ hashFiles('deps/WORKSPACE.bazel.tpl', 'deps/BUILD.bazel') }}
          path: ~/.cache/bazel
          restore-keys: |
            bazel-cache.ray-jl.${{ matrix.os }}.${{ matrix.arch }}.julia-${{ matrix.version }}.ray-${{ steps.ray-commit.outputs.head_sha }}.
            bazel-cache.ray-jl.${{ matrix.os }}.${{ matrix.arch }}.julia-${{ matrix.version }}.
            bazel-cache.ray-jl.${{ matrix.os }}.${{ matrix.arch }}.
```

Ray.jl/test/runtests.jl
```diff
@@ -0,0 +1,9 @@
+include("set_up_tests.jl")
```
omus, line 1: Better name? Is there a precedent for this?
```suggestion
include("import_deps.jl")
```

.github/workflows/Ray_CI.yml
kleinschmidt: eh, given that we're gonna split this out I think this is better (to make that separation less painful)

.github/workflows/Ray_CI.yml
kleinschmidt: ah that's an interesting point.  I'd hoped to actually re-use the jll cache; but i guess if we don't need to do _full_ re-builds then it's probably okay.

Ray.jl/test/runtests.jl
kleinschmidt: yeah, this is a beacon package templates thing to make running tests locally less painful.  but I'm not sure it's needed here (no juggling AWS permissiosn or anything of the sort)

Ray.jl/docs/fix_doctests.jl
```diff
@@ -0,0 +1,20 @@
...
+# they are correct (and be easily revertable if they do something wrong).
+
+using Documenter, Ray
+
+DocMeta.setdocmeta!(Ray, :DocTestSetup, :(using Ray);
+                    recursive=true)
+
+if get(ENV, "CI", "false") == "true" || success(`git diff --quiet`)
+    if ismissing(get(ENV, "AWS_PROFILE", missing))
+        @warn """You may need to set `ENV["AWS_PROFILE"] = ray_core_worker_julia_jll-ci` in order to successfully run the doctests"""
```
glennmoy, line 15: do we need to make this role (or similar)?

Ray.jl/docs/make.jl
```diff
@@ -0,0 +1,13 @@
...
+makedocs(; modules=[Ray],
+         sitename="Ray.jl",
+         authors="Beacon Biosignals",
+         pages=["API Documentation" => "index.md"],
+         strict=true)
+
+deploydocs(; repo="github.com/beacon-biosignals/ray_core_worker_julia_jll.jl.git", push_preview=true,
+           devbranch="main", dirname="Ray",
+           versions=["stable" => "Ray-v^", "Ray-v#.#",
+                     "dev" => "dev"])
```
glennmoy, line 13:
```suggestion
                     "dev" => "main"])
```
I assume?

Ray.jl/docs/Project.toml
```diff
@@ -0,0 +1,6 @@
+[deps]
```
glennmoy, line 1: isn't there normally a Manifest for the docs as well?

glennmoy (approved):

.github/workflows/Ray_CI.yml
```diff
@@ -0,0 +1,93 @@
+name: Ray.jl - CI
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    tags: ["*"]
+    paths: ["Ray.jl/**", ".github/workflows/Ray_CI.yml"]
```
omus, line 8:
```suggestion
    paths:
      - "Ray.jl/**"
      - ".github/workflows/Ray_CI.yml"
```

.github/workflows/Ray_CI.yml
```diff
@@ -0,0 +1,93 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths: ["Ray.jl/**",".github/workflows/Ray_CI.yml"]
```
omus, line 11:
```suggestion
    paths:
      - "Ray.jl/**"
      - ".github/workflows/Ray_CI.yml"
```

Ray.jl/docs/fix_doctests.jl
kleinschmidt: I should just remove all teh docs stuff for now, we're not using it and this is from a template for a Lab sub-package which doesn't really map well onto this

Ray.jl/docs/make.jl
kleinschmidt: it's the sub-directory where dev docs are built (only have to specify this if we wanna deploy docs in this sub-package).  added a breadcrumb to change these settings when we move to own repo

Ray.jl/docs/Project.toml
kleinschmidt: I just nuked all the docs stuff sicne it doesn't make sense trying to build them right now (all the caching etc. would have to be set up in teh docs build)

Ray.jl/test/runtests.jl
kleinschmidt: I removed this in #19

PR #19, opened 2023-08-10
function manager prototype

kleinschmidt:
very WIP at the moment.  needs a real worker to coordinate the job id etc. but this is the basic proof of concept!


kleinschmidt:
Okay, I'm doing a bit more reading on the situation with global refs + serialization.  Apparently, there's a special serializer (ClusterSerializer) which inspects the IR of anonymous functions, detects global refs, and sends them to the remote nodes when the function is serialized.  I really would not like to go down that road, since it's going to make function import non-stateful (unless we define the function in generated module where we can also define the global refs it needs, but that probably would lead to all kinds of other complications).

However, I worry that _without_ supporting global ref handling, our users will have to be very careful to actually capture all necessary values when defining their functions to be executed as tasks, by either
1. defining "work" functions in a package which provides an environment that can be instantiated via "normal" code loading/dependency resolution mechanisms
2. passing everything not imported from a dependency as an argument to the function
3. defining work functions in a local scope (let block, funciton factory) so that a "real closure" is returned (struct with properties holding closed-over bindings)

even so, 3. is still a bit of a footgun because if there's anything that's _not_ captured in that local scope, you'll still get undef errors on the remote.

another thing we can consider doing is at least trying to detect global refs in function IR to be serialized (can maybe even use the machinery from clusterserializer) and throwing an informative error if any are found.

kleinschmidt:
Some histoyr: https://github.com/JuliaLang/julia/pull/19594/files

kleinschmidt:
CI is failing I think due to race conditions, but that's just a hunch.

kleinschmidt:
test failure on 1.8 is real and due somehow to the fact that the tests are running inside a function body/local scope.  I can reproduce locally by doing

```julia
let
    include("function_manager.jl")
end
```

after running the setup code for the tests.

The error suggests that there's some deeep weirdness happening with this deserialization:

```
  Got exception outside of a @test
  On worker 2:
  UndefVarError: M not defined
  Stacktrace:
    [1] deserialize_module
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:996
    [2] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:895
    [3] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:813
    [4] deserialize_datatype
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:1363
    [5] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:866
    [6] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:813
    [7] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:873
    [8] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:813 [inlined]
    [9] deserialize_dict
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:1495
   [10] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:1502
   [11] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:882
   [12] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:1479
   [13] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:877
   [14] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:1467
   [15] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:959
   [16] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:1479
   [17] handle_deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:877
   [18] deserialize
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Serialization/src/Serialization.jl:813 [inlined]
   [19] deserialize_msg
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Distributed/src/messages.jl:87
   [20] #invokelatest#2
      @ ./essentials.jl:729 [inlined]
   [21] invokelatest
      @ ./essentials.jl:726 [inlined]
   [22] message_handler_loop
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Distributed/src/process_messages.jl:176
   [23] process_tcp_streams
      @ /opt/hostedtoolcache/julia/1.8.5/x64/share/julia/stdlib/v1.8/Distributed/src/process_messages.jl:133
   [24] #103
      @ ./task.jl:484
```

kleinschmidt:
Ahh, I think I'm starting to understand what's going on: it's the [_thunk_ passed to remotecall_fetch](https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/pull/19/files#diff-b46d6263f6fb5780ebe899ebdcf2b03f5f7aaefdc0aa8c3eb7151db442340d63R65-R74) that's not getting deserialized properly, _not_ the exported function.  (How do I know this?  I tried putting in some printlns inside the remotecall'd function and they never get printed at all)

That is, this is an issue with how the test suite is setup, NOT the actual functionality being tested.

kleinschmidt:
Okay, I think this is actually due to something to do with the scope introduced by `@testset`.  Not sure there's really any way to work around that for these tests so I think I may just 🔪 the remote worker tests...

kleinschmidt:
I have no idea why the funciton store is timing out only on 1.9

glennmoy:
Just to point out that Ray advises against passing around global variables to remote functions: https://docs.ray.io/en/latest/ray-core/patterns/global-variables.html#anti-pattern-using-global-variables-to-share-state-between-tasks-and-actors

So _not_ supporting this (or punting on it) wouldn't be the end of the world IMO.

Ray.jl/src/function_manager.jl
```diff
@@ -0,0 +1,78 @@
...
+        @debug "retrieving $(fd) from function store with key $(key)"
+        val = Get(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, -1)
+        try
+            io = IOBuffer()
+            iob64 = Base64DecodePipe(io)
+            write(io, val)
+            seekstart(io)
+            f = deserialize(iob64)
+        catch e
+            error("Failed to deserialize function from store: $(fd)")
```
glennmoy, line 75: maybe rethrow the error?

Ray.jl/src/function_manager.jl
```diff
@@ -0,0 +1,78 @@
...
+    if Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, -1)
+        @debug "function already present in GCS store:" key f
+    else
+        @debug "exporting function to GCS store:" key f
+        val = base64encode(serialize, f)
+        Put(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, val, true, -1)
+    end
+end
+
+function wait_for_function(fm::FunctionManager, fd::JuliaFunctionDescriptor, job_id;
```
glennmoy, line 51: where do we expect to use this?

Ray.jl/test/function_manager.jl
```diff
@@ -0,0 +1,84 @@
...
+    fd = function_descriptor(f)
+    @test wait_for_function(fm, fd, jobid) == :ok
+    f2 = import_function!(fm, fd, jobid)
+
+    @test f2.(1:10) == f.(1:10)
+
+    mfd = function_descriptor(MyMod.f)
+    @test_throws ErrorException import_function!(fm, mfd, jobid)
+    @test wait_for_function(fm, mfd, jobid; timeout_s=1) == :timed_out
+    export_function!(fm, MyMod.f, jobid)
```
glennmoy, line 24: what's this testing exactly? that `MyMod.f` won't be confused with `f` in the function manager?

Ray.jl/src/function_manager.jl
```diff
@@ -0,0 +1,78 @@
...
+end
+
+# TODO: remove indirection after fields are settled
+FunctionManager = FunctionManager1
+
+# XXX: we should probably be packaging task-related metadata like the job_id in
+# a task/remotefunction-like struct instead of passing it around like this.
+function_key(fd::JuliaFunctionDescriptor, job_id) = string("RemoteFunction:", job_id, ":", fd.function_hash)
+
+function export_function!(fm::FunctionManager, f, job_id)
```
glennmoy, line 39: we can bikeshed on this later but I'm wondering if we should overload `setproperty!` and `getproprty` for the FunctionManager.

Ray.jl/test/function_manager.jl
```diff
@@ -0,0 +1,84 @@
...
+    Connect(client)
+
+    fm = FunctionManager(client, Dict{String,Any}())
+
+    jobid = 1337
+
+    f = x -> isless(x, 5)
+    export_function!(fm, f, jobid)
+
+    fd = function_descriptor(f)
```
glennmoy, line 15: I'd find it helpful to know the functionality being tested if we organised these into`@testsets`

Ray.jl/src/function_manager.jl
kleinschmidt: I added it so that I can do a timed wait in the tests; I expect to use it on workers when there may be a race condition vis a vis the functions get exported adn the task being submitted...you don't want to fail out of task execution just because the function hasn't propagated yet.

kleinschmidt: tests are still failing though so my hypothesis that it was a race condition is clearly wrong lol https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/actions/runs/5860355101/job/15888248877?pr=19#step:11:167

kleinschmidt:
okay, I guess this is fine.  CI seems to pass as long as there's _some kind_ of reference ot the function descriptor further down even if it's a debug statemetn that never actually gets printed.  I have no idea why holding onto the _function descriptor_ would affect whether or not a string that 's constructed from one of its fields gets mangled but here we are.

Ray.jl/src/function_manager.jl
```diff
@@ -0,0 +1,78 @@
...
+    fd = function_descriptor(f)
+    key = function_key(fd, job_id)
+    if Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, -1)
+        @debug "function already present in GCS store:" key f
+    else
+        @debug "exporting function to GCS store:" key f
+        val = base64encode(serialize, f)
```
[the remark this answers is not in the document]
kleinschmidt: yeah, because serialized julia functions have lots of null bytes `"\0"` in them.  when you pass them directly to methods that expect a `std::string` it truncates the string at the first null byte (since the default constructor on the C++ side takes a C string, which is null-terminated).

there _are_ ways to construct a `std::string` directly with null bytes but apparently CxxWrap doesn't support htem, at least based on my very cursory poking around::

```
julia> StdString(Int8[1, 0, 2])
"\x01"

julia> StdString(Int8[1, 0, 2], 3)
"\x01"
```

Ray.jl/test/runtests.jl
```diff
@@ -1,9 +1,34 @@
...
+module MySubMod
+f(x) = x - 1
+end # module MM
+
+end # module M
+
+module NotMyMod
+using ..MyMod: f
+g(x) = x - 1
+end
```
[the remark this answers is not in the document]
kleinschmidt: I sswear to god if this upsets my teetering house of cards CI I'm going to fully lose it

Ray.jl/src/function_manager.jl
glennmoy: if it's only expected to be used in tests would it be better placed in `runtests.jl`? If we expect to use it in the main source code then ignore.

glennmoy (approved):

Ray.jl/src/function_manager.jl
kleinschmidt: ultimately we _will_ use it in actual code so I think it shoudl stay here (it's based on a similar thing on the python side...)

test/runtests.jl
```diff
@@ -1,11 +1,15 @@
 using CxxWrap
 using Test
-using ray_core_worker_julia_jll: FunctionDescriptor, function_descriptor
+using ray_core_worker_julia_jll: JuliaFunctionDescriptor, function_descriptor
 using ray_core_worker_julia_jll: initialize_coreworker, shutdown_coreworker
 using ray_core_worker_julia_jll: get, put, submit_task
 
 include("utils.jl")
 
+module M
```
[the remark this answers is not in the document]
kleinschmidt: did something similar already: https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/pull/19/files#diff-ed0de87b16d7f62b2793722f7901380bb28bf3d1abd687a172f51cfc346fb58fR1-R14

PR #20, opened 2023-08-11
Execute a predefined named Julia function as a Ray task

omus:
Allows for executing arbitrary Julia functions with these restrictions:

- The function must accessible when loading a fresh Julia session with no imports
- Function must take no arguments
- Function must return an `Int32`

The primary progress here is that the specified function is stored in a function descriptor which is passed to the `task_executor` function in Julia.

### DIFF
--- a/src/wrappers/any.jl
+++ b/src/wrappers/any.jl
@@ -124,8 +124,9 @@ function Base.take!(buffer::CxxWrap.CxxWrapCore.SmartPointer{<:Buffer})
     return vec
 end
 
-function task_executor()
+function task_executor(ray_function)
     @info "task_executor called"
+    @info "lang is julia: $(GetLanguage(ray_function) == Language.JULIA)"
     return getpid()
 end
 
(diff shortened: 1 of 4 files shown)


src/wrappers/any.jl
```diff
@@ -124,14 +124,20 @@ function Base.take!(buffer::CxxWrap.CxxWrapCore.SmartPointer{<:Buffer})
     return vec
 end
 
-function task_executor()
+function task_executor(ray_function)
     @info "task_executor called"
-    return getpid()
+    fd = GetFunctionDescriptor(ray_function)
+    func = eval(@__MODULE__, Meta.parse(CallString(fd)))
```
omus, line 130: Probably makes more sense to use `Main` here. Needs to be some extra thought

src/wrappers/any.jl
```diff
@@ -190,12 +196,27 @@ function start_worker(args=ARGS)
...
-        @info "Testing"
-        initialize_coreworker_worker(
-            parsed_args["node_manager_port"],
-            CxxWrap.@safe_cfunction(task_executor, Int32, ()),
-        )
-    end
+    # Note (omus): Logging is currently limited to a single worker as all workers attempt to
+    # write to the same file.
+    global_logger(FileLogger(joinpath(parsed_args["logs_dir"], "julia_worker.log");
+                             append=true, always_flush=true))
```
kleinschmidt, line 202: I needed to add `FileLogger` (from LoggingExtras.jl`) here otherwise the file IO doens't flush on each write and you get no logs at all

src/wrappers/any.jl
kleinschmidt: yeah I think that's basically right...this is the callback that the (worker) coreworker calls when a task is scheduled to execute on that worker, which hands control over to the julia runtime to actually do the work.

kleinschmidt (approved):

PR #24, opened 2023-08-15
pull worker marshalling functions into Ray.jl

kleinschmidt:
This is necessary in order to use the function manager in the task executor callback, since it's defined in Ray.jl and not in the jll.

The alternative would be to define the function manager in the jll, but I think this is better: IMO the jll should basically _only_ have the wrapped C++ types/functions and whatever minimal julia code is required to expose them in a meaningful way.  the julia code that manages the runtime is pretty complex and likely to become more so as we support more features, so keeping it in Ray.jl is better.

The one wart so far here is that we now have to `@eval` the `@safe_cfunction` invocation because `task_executor` is not known are compile time (and can't be, since it's now defined in Ray.jl).  The alternative would be to do the `@safe_cfunction` in Ray.jl itself, but I'm trying really hard to avoid having direct CxxWrap nonsense in Ray.jl.

Depends on https://github.com/beacon-biosignals/ray/pull/6 for task test to pass (they do pass locally)

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -12,8 +12,26 @@ function setup_ray_head_node(body)
     end
 end
 
+function node_manager_port()
+    line = open("/tmp/ray/session_latest/logs/raylet.out") do io
+        while !eof(io)
+            line = readline(io)
+            if contains(line, "NodeManager server started")
+                return line
+            end
+        end
+    end
+
+    m = match(r"port (\d+)", line)
+    return m !== nothing ? parse(Int, m[1]) : error("Unable to find port")
+end
+
 function setup_core_worker(body)
-    initialize_coreworker()
+    initialize_coreworker("/tmp/ray/session_latest/sockets/raylet",
+                          "/tmp/ray/session_latest/sockets/plasma_store",
+                          "127.0.0.1:6379",
+                          "127.0.0.1",
+                          node_manager_port())
     try
         body()
     finally


Ray.jl/src/Ray.jl
```diff
@@ -5,9 +5,19 @@ This package provides user-facing interface for Julia-on-Ray.
...
+using LoggingExtras
+using Pkg
 using Serialization
-using ray_core_worker_julia_jll
+
+using ray_core_worker_julia_jll: shutdown_coreworker
+
+import ray_core_worker_julia_jll as rayjll
+
+export start_worker, initialize_coreworker, shutdown_coreworker, submit_task
```
glennmoy, line 19:
```suggestion
export initialize_coreworker, shutdown_coreworker
export start_worker, submit_task
```

Ray.jl/src/runtime.jl
```diff
@@ -0,0 +1,145 @@
+function bootstrap_ray_args_from_raylet_out()
```
glennmoy, line 1:
```suggestion
function parse_ray_args_from_raylet_out()
```

not fussed about the semantics tbh

Ray.jl/src/runtime.jl
```diff
@@ -0,0 +1,145 @@
...
+
+    # TODO: downgrade to debug
+    @info "Raylet socket: $raylet, Object store: $store, Node IP: $node_ip, Node port: $node_port, GCS Address: $gcs_address"
+
+    return (raylet, store, gcs_address, node_ip, node_port)
+end
+
+# TODO: use something like the java config bootstrap address (?) to get this
+# information instead of parsing logs?  I can't quite tell where it's coming
+# from (set from a `ray.address` config option):
```
glennmoy, line 52: yeah it'd be nice if these were env variables or something

Ray.jl/src/runtime.jl
```diff
@@ -0,0 +1,145 @@
...
+# https://github.com/beacon-biosignals/ray/blob/beacon-main/java/runtime/src/main/java/io/ray/runtime/config/RayConfig.java#L165-L171
+initialize_coreworker() = initialize_coreworker(bootstrap_ray_args_from_raylet_out()...)
+initialize_coreworker(args...) = rayjll.initialize_coreworker(args...)
```
glennmoy, line 55:
```suggestion
initialize_coreworker() = rayjll.initialize_coreworker(bootstrap_ray_args_from_raylet_out()...)
```

unless we need the method with `args`?

Ray.jl/src/runtime.jl
```diff
@@ -0,0 +1,145 @@
...
+            arg_type=Int
+            default=0
+            help="The computed hash of the runtime env for this worker"
+        "arg1"
+            required = true
+    end
+
+    parsed_args = parse_args(args, s)
+
+    logfile = joinpath(parsed_args["logs_dir"], "julia_worker_$(getpid()).log")
```
glennmoy, line 133: good idea

glennmoy (approved):
LGTM

Ray.jl/src/runtime.jl
kleinschmidt: yeah good question; I thnk we probably _don't_ need it for now.

PR #21, opened 2023-08-14
Parse CoreWorker options from Ray cli passthrough arguments

glennmoy:
Currently we hardcode the `CoreWorkerOptions` values in both `initialize_coreworker` and `initialize_coreworker_worker` functions. 

These changes allow for `initialize_coreworker_worker` to parse these values from the ray CLI `passthrough` args.

For the `initialize_coreworker` function, we continue to parse the `raylet.out` logs for CLI args.

~~The `raylet_socket` and `store_socket` continue to be hardcoded but I'm almost certain these are sensible defaults anyway.~~

PR #25, opened 2023-08-15
integrate function manager and task prototypes

kleinschmidt:
This is a followup to #20 and #19 which 
- introduces a global function manager in Ray.jl
- slightly improves initialization and jobid handling (now gets jobid from coreworkerprocess)
- exports function on task submit
- imports function in task executor

I'd say it's still WIP; a _single_ task invocation works great!  even with `submit_task(() -> getpid())`.  but for some reaosn things deadlock on any further task submissions, with no errors in the raylet.err or `julia_worker_$(PID).log` files.

### DIFF
--- a/Ray.jl/src/function_manager.jl
+++ b/Ray.jl/src/function_manager.jl
@@ -38,6 +38,14 @@ end
 
 const FUNCTION_MANAGER = Ref{FunctionManager}()
 
+function _init_global_function_manager(gcs_address)
+    @info "connecting function manager to GCS at $gcs_address..."
+    gcs_client = JuliaGcsClient(gcs_address)
+    rayjll.Connect(gcs_client)
+    FUNCTION_MANAGER[] = FunctionManager(; gcs_client,
+                                         functions=Dict{String,Any}())
+end
+
 function function_key(fd::JuliaFunctionDescriptor, job_id=get_current_job_id())
     return string("RemoteFunction:", job_id, ":", fd.function_hash)
 end
@@ -45,6 +53,7 @@ end
 function export_function!(fm::FunctionManager, f, job_id=get_current_job_id())
     fd = function_descriptor(f)
     key = function_key(fd, job_id)
+    @debug "exporting function to function store:" fd key
     if Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE,
               deepcopy(key), # DFK: I _think_ the string memory may be mangled
                              # if we don't copy.  not sure but it can't hurt
@@ -70,11 +79,14 @@ function wait_for_function(fm::FunctionManager, fd::JuliaFunctionDescriptor,
 end
 
 # XXX: this will error if the function is not found in the store.
+# TODO: consider _trying_ to resolve the function descriptor locally (i.e.,
+# somthing like `eval(Meta.parse(CallString(fd)))`), falling back to the function
+# store only if needed.
 function import_function!(fm::FunctionManager, fd::JuliaFunctionDescriptor,
                           job_id=get_current_job_id())
     return get!(fm.functions, fd.function_hash) do
         key = function_key(fd, job_id)
-        @debug "retrieving $(fd) from function store with key $(key)"
+        @debug "function not found locally, retrieving from function store" fd key
         val = Get(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, -1)
         try
             io = IOBuffer()
@@ -82,6 +94,9 @@ function import_function!(fm::FunctionManager, fd::JuliaFunctionDescriptor,
             write(io, val)
             seekstart(io)
             f = deserialize(iob64)
+            # need to handle world-age issues on remote workers when
+            # deserializing the function effectively defines it
+            return (args...) -> invokelatest(f, args...)
         catch e
             error("Failed to deserialize function from store: $(fd)")
         end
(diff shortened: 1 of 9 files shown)


kleinschmidt:
I see the same deadlock on repeated task submission on main so I'm not _too_ fussed about it vis a vis this PR, but we obviously need to understand why that's happening and fix it.

kleinschmidt:
Did a bit more debugging and see this deadlock-on-second-task behavior even on `74abc3a @ Parse CoreWorker options from Ray cli passthrough arguments (#21)`  so I'm pretty sure it's not something introduced in #24.  I'm gonna mark this ready for review then.

Ray.jl/src/runtime.jl
```diff
@@ -159,6 +161,9 @@ function start_worker(args=ARGS)
 
     parsed_args = parse_args(args, s)
 
+    _init_global_function_manager(parsed_args["address"])
+
+    ENV["JULIA_DEBUG"] = "Ray"
```
glennmoy, line 166: should this go at the top level of the module?

Ray.jl/src/runtime.jl
```diff
@@ -1,3 +1,28 @@
...
+    # initialization and using that as a proxy for whether it's connected or not
+    #
+    # https://github.com/beacon-biosignals/ray/blob/dfk%2Fusing-Ray/python/ray/_private/worker.py#L421
+    if isassigned(FUNCTION_MANAGER)
+        @warn "Ray already initialized, skipping..."
+        return nothing
+    end
+
+    args = parse_ray_args_from_raylet_out()
+    initialize_coreworker(args...)
```
glennmoy, line 13: for the future - but I'd like to rename this as `initialize_driver`

Ray.jl/src/function_manager.jl
```diff
@@ -22,25 +22,38 @@
...
 
 Base.@kwdef struct FunctionManager
     gcs_client::JuliaGcsClient
     functions::Dict{String,Any}
 end
 
-# XXX: we should probably be packaging task-related metadata like the job_id in
-# a task/remotefunction-like struct instead of passing it around like this.
-function_key(fd::JuliaFunctionDescriptor, job_id) = string("RemoteFunction:", job_id, ":", fd.function_hash)
+const FUNCTION_MANAGER = Ref{FunctionManager}()
```
glennmoy, line 39: this is only accessible to coreworkers that share memory right?

Ray.jl/src/runtime.jl
```diff
@@ -1,3 +1,28 @@
+function init()
+    # XXX: this is at best EXREMELY IMPERFECT check.  we should do something
+    # more like what hte python Worker class does, getting node ID at
+    # initialization and using that as a proxy for whether it's connected or not
+    #
+    # https://github.com/beacon-biosignals/ray/blob/dfk%2Fusing-Ray/python/ray/_private/worker.py#L421
```
glennmoy, line 6: this gives me a 404 - I think you linked to you own branch?

Ray.jl/src/function_manager.jl
```diff
@@ -22,25 +22,38 @@
...
-# maybe job id?
+# ~~maybe job id?~~ this is managed by the core worker process
 
-using ray_core_worker_julia_jll: JuliaGcsClient, Exists, Put, Get, JuliaFunctionDescriptor, function_descriptor
+using ray_core_worker_julia_jll: JuliaGcsClient, Exists, Put, Get,
+                                 JuliaFunctionDescriptor, function_descriptor
 
-# XXX: what's the actual namespace to use?  probably set per-job but I dunno.
-const FUNCTION_MANAGER_NAMESPACE = "JuliaFunctions"
+# python uses "fun" for the namespace: https://github.com/beacon-biosignals/ray/blob/dfk%2Fusing-Ray/python/ray/_private/ray_constants.py#L380
```
glennmoy, line 30: github link is to your own branch

deps/wrapper.cc
```diff
@@ -13,6 +13,7 @@ void initialize_coreworker(
     options.language = Language::JULIA;
     options.store_socket = store_socket;    // Required around `CoreWorkerClientPool` creation
     options.raylet_socket = raylet_socket;  // Required by `RayletClient`
+    // XXX: this is hard coded! very bad!!!
```
glennmoy, line 16: I think there some method `JobID::FromNext()` we can use?

Ray.jl/src/runtime.jl
kleinschmidt: probably shouldn't even be here at all (shoudl have a switch or something to turn on debug logging that gets passed in the runtime env) but was the most direct way to do it on the workers

Ray.jl/src/function_manager.jl
kleinschmidt: yeah, so running under the same process.  may need to protect it with a lock or something (or build the lock into the function manager itself) in case of any async/multithreading shenanigans but I think this is fine for now

Ray.jl/src/runtime.jl
kleinschmidt: changed all these to use the git commit hash instead of branch name (and changed my editor config to generate those from now on 😁 )

deps/wrapper.cc
kleinschmidt: or just call it from the cpp init function but that seems a bit wasteful to me somehow.  maybe we don't really need to track state on the julia side though?

PR #26, opened 2023-08-17
use void pointer casting to pass args 

kleinschmidt:
I dunno if this is a good idea or not but it works for me locally.  Unfortunately had to punt on the safe cfunction since I was getting some weird CxxWrap errors about no julia wrapper for <mangled name> (very short so I'm assuming void pointer related).

but using cfunction directly I can at least get it to take the void pointer to the args list and static cast it in the callback executor to the actual list.


kleinschmidt:
closing in favor of https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/pull/23/commits/7ea719e3d36037f5a399e27195aa85b1590dc64d

PR #28, opened 2023-08-18
get jobid from GCS global state accessor

kleinschmidt:
I ended up removing the no-argument `initialize_coreworker_driver` entrypoint since I didn't want to put the job id detection stuff in there (rather than in `init`), but not 100% wedded to that decision.

for now the only method I wrapped from the global state accessor is the one to get the next job id (plus setup/teardown stuff).  I don't _think_ we really need any more than that.  In fact, the vanilla GCS client has a next job ID endpoint as well, but I considered and rejected that in #16 (in favor of hte "python gcs client" which doesn't like kill your session when it fails to connect or require fiddling with threads/asyncio/boost).

### DIFF
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -1,3 +1,5 @@
+const GLOBAL_STATE_ACCESSOR = Ref{rayjll.GlobalStateAccessor}()
+
 function init()
     # XXX: this is at best EXREMELY IMPERFECT check.  we should do something
     # more like what hte python Worker class does, getting node ID at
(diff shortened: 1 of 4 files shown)


Ray.jl/src/runtime.jl
```diff
@@ -1,3 +1,5 @@
+const GLOBAL_STATE_ACCESSOR = Ref{rayjll.GlobalStateAccessor}()
```
glennmoy, line 1: can we attach a docstring to this just to explain what it is?

glennmoy (approved):

PR #31, opened 2023-08-18
Enable single arbitrary return value from task_executor

glennmoy:
---
  - https://app.asana.com/0/0/1205303690266611


glennmoy:
Getting an error about undefined julia wrapper type, I think I know the drill on how to fix these now...

```julia
[ Info: connecting function manager to GCS at 127.0.0.1:6379...
ERROR: Type NSt3__16vectorINS_4pairIN3ray8ObjectIDENS_10shared_ptrINS2_9RayObjectEEEEENS_9allocatorIS7_EEEE has no Julia wrapper
Stacktrace:
 [1] initialize_coreworker_worker(arg1::AbstractString, arg2::AbstractString, arg3::AbstractString, arg4::AbstractString, arg5::Integer, arg6::CxxWrap.CxxWrapCore.SafeCFunction)
 [2] start_worker(args::Vector{String})
   @ Ray ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/runtime.jl:204
 [3] start_worker()
   @ Ray ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/runtime.jl:148
```

glennmoy:
It seems `std::pair` isn't wrapped by CxxWrap - advice is to use `std::tuple`
https://github.com/JuliaInterop/CxxWrap.jl/issues/201

Ray.jl/test/task.jl
```diff
@@ -6,7 +6,7 @@
 
     addme(x...) = Int32(sum(x))
     oid = submit_task(addme, 1, 2, 3)
-    result = String(take!(ray_core_worker_julia_jll.get(oid)))
+    result = take!(ray_core_worker_julia_jll.get(oid))
```
glennmoy, line 9: this test will obviously fail until we extract the data from the RayObject

glennmoy: running this locally the segfault occurs before we `take!`

src/wrappers/any.jl
```diff
@@ -146,6 +146,15 @@ end
 # CxxWrap as a direct dependency in Ray.jl
 _submit_task(dir, fd, oids::AbstractVector) = _submit_task(dir, fd, StdVector(oids))
 
+function RayObject(x)
+    io = IOBuffer()
+    serialize(io, x)
+    data_buffer = LocalMemoryBuffer(Ptr{Nothing}(pointer(io.data)), sizeof(io.data), true)
+    meta_buffer = LocalMemoryBuffer(Ptr{Nothing}(), 0, false)
```
glennmoy, line 153: I also tried `LocalMemoryBuffer(CxxPtr{UInt8}(C_NULL), 0, false)`

deps/wrapper.cc
```diff
@@ -81,12 +82,10 @@ void initialize_coreworker_worker(
...
-          // task_executor(ray_function, returns, args);
-          int pid = task_executor(ray_function, args);
-          std::string str = std::to_string(pid);
-          auto memory_buffer = std::make_shared<LocalMemoryBuffer>(reinterpret_cast<uint8_t *>(&str[0]), str.size(), true);
+          RayObject result = task_executor(ray_function, args);
+
           RAY_CHECK(returns->size() == 1);
-          (*returns)[0].second = std::make_shared<RayObject>(memory_buffer, nullptr, std::vector<rpc::ObjectReference>());
+          (*returns)[0].second = std::make_shared<RayObject>(std::move(result));
           return Status::OK();
```
glennmoy, line 89: printing debug statements shows the segfault doesn't happen until after the julia task returns
```julia
[2023-08-21 13:34:08,779 I 4983 3209946] wrapper.cc:85: Task Executing... 
[2023-08-21 13:34:09,588 I 4983 3209946] wrapper.cc:87: Task Finished... 
[2023-08-21 13:34:09,588 I 4983 3209946] wrapper.cc:90: Appending returns... 
[2023-08-21 13:34:09,588 I 4983 3209946] wrapper.cc:92: STATUS::OK 

[4983] signal (11.2): Segmentation fault: 11
in expression starting at none:1
unknown function (ip: 0xa9027bfda9014ff4)
Allocations: 5938043 (Pool: 5932656; Big: 5387); GC: 8
[2023-08-21 13:34:11,023 E 4978 3209873] (raylet) agent_manager.cc:135: The raylet exited immediately because the Ray agent failed. The raylet fate shares with the agent. This can happen because the Ray agent was unexpectedly killed or failed. Agent can fail when
- The version of `grpcio` doesn't follow Ray's requirement. Agent can segfault with the incorrect `grpcio` version. Check the grpcio version `pip freeze | grep grpcio`.
- The agent failed to start because of unexpected error or port conflict. Read the log `cat /tmp/ray/session_latest/logs/dashboard_agent.log`. You can find the log file structure here https://docs.ray.io/en/master/ray-observability/ray-logging.html#logging-directory-structure.
- The agent is killed by the OS (e.g., out of memory).
```

Ray.jl/src/runtime.jl
```diff
@@ -131,9 +134,20 @@ function task_executor(ray_function, ray_objects)
...
+
+    # TODO: remove - useful for now for debugging
+    @info "Result: $result"
+
+    # TODO: support multiple return values
+    buffer_data = Vector{UInt8}(sprint(serialize, result))
+    buffer_size = sizeof(buffer_data)
+    buffer = rayjll.LocalMemoryBuffer(buffer_data, buffer_size, true)
+    push!(returns, buffer)
+    return nothing
```
omus, line 150:
```suggestion

    return nothing
```

Ray.jl/test/task.jl
```diff
@@ -1,13 +1,12 @@
 @testset "Submit task" begin
-    oid1 = submit_task(Int32 ∘ length, "hello")
-    addme(x...) = Int32(sum(x))
-    oid2 = submit_task(addme, 1, 2, 3)
+    oid1 = submit_task(length, "hello")
+    oid2 = submit_task(max, 1, 2)
```
omus, line 3: Slightly more interesting test which results in a non-`Int` return type:
```suggestion
    oid2 = submit_task(max, 0x00, 0xff)
```

Ray.jl/test/task.jl
```diff
@@ -1,13 +1,12 @@
...
+    result1 = deserialize(IOBuffer(take!(ray_core_worker_julia_jll.get(oid1))))
+    @test result1 isa Int
+    @test result1 == 5
 
-    result2 = String(take!(ray_core_worker_julia_jll.get(oid2)))
-    @test all(isdigit, result2)
-    @test parse(Int, result2) == 6
+    result2 = deserialize(IOBuffer(take!(ray_core_worker_julia_jll.get(oid2))))
+    @test result2 isa Int
+    @test result2 == 2
```
omus, line 11:
```suggestion
    @test result2 isa UInt8
    @test result2 == 0xff
```

deps/wrapper.cc
```diff
@@ -42,12 +42,14 @@ void initialize_coreworker_worker(
...
     int64_t startup_token,
-    jlcxx::SafeCFunction julia_task_executor) {
-    auto task_executor = jlcxx::make_function_pointer<int(
-        RayFunction,
-        std::vector<std::shared_ptr<RayObject>>
-        // std::vector<std::pair<ObjectID, std::shared_ptr<RayObject>>> *returns
-    )>(julia_task_executor);
+    void *julia_task_executor) {
+
+    // XXX: Ideally the task_executor would take the expected callback arg types:
```
omus, line 47:
```suggestion
    // XXX: Ideally the task_executor would use a `jlcxx::SafeCFunction` and take the expected
    // callback arg types:
```

deps/wrapper.cc
```diff
@@ -82,13 +84,17 @@ void initialize_coreworker_worker(
...
+          auto args_ptr = static_cast<const void *>(&args);
+          task_executor(ray_function, &return_vec, args_ptr);
+
+          RAY_CHECK(return_vec.size() == 1);
           RAY_CHECK(returns->size() == 1);
-          (*returns)[0].second = std::make_shared<RayObject>(memory_buffer, nullptr, std::vector<rpc::ObjectReference>());
+
+          // TODO: support multiple return values
+          std::shared_ptr<LocalMemoryBuffer> buffer = return_vec[0];
+          (*returns)[0].second = std::make_shared<RayObject>(buffer, nullptr, std::vector<rpc::ObjectReference>(), true);
```
omus, line 97: Can we get away with using `copy=false`?
```suggestion
          (*returns)[0].second = std::make_shared<RayObject>(buffer, nullptr, std::vector<rpc::ObjectReference>(), false);
```

deps/wrapper.cc
```diff
@@ -373,6 +389,7 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
     mod.method("LocalMemoryBuffer", [] (uint8_t *data, size_t size, bool copy_data = false) {
         return std::make_shared<LocalMemoryBuffer>(data, size, copy_data);
     });
+    jlcxx::stl::apply_stl<std::shared_ptr<Buffer>>(mod);
```
omus, line 392: Move the `std::vector` support to be next to `Buffer`
```suggestion
    jlcxx::stl::apply_stl<std::shared_ptr<Buffer>>(mod);
    mod.method("LocalMemoryBuffer", [] (uint8_t *data, size_t size, bool copy_data = false) {
        return std::make_shared<LocalMemoryBuffer>(data, size, copy_data);
    });
```

Ray.jl/src/runtime.jl
```diff
@@ -115,7 +115,10 @@ function submit_task(f::Function, args...)
     return GC.@preserve args rayjll._submit_task(project_dir(), fd, object_ids)
 end
 
-function task_executor(ray_function, ray_objects)
+function task_executor(ray_function, returns, ray_objects)
+    returns = rayjll.cast_to_buffer(returns)
+    ray_objects = rayjll.cast_to_rayobject(ray_objects)
```
omus, line 120: These function names are rather inaccurately named. Would it be better to do:
```suggestion
    returns = rayjll.cast_to_returns(returns)
    ray_objects = rayjll.cast_to_arg_objects(ray_objects)
```
Maybe we should rename `ray_objects` to `arg_objects`? The name `ray_objects` doesn't indicate args at all.

glennmoy: yeah I'm fine with `arg_objects`. But I reworded `cast_to_returns` to `cast_to_buffer` because it's only incidental to the fact that it's used for "task returns" in the end.

src/wrappers/any.jl
```diff
@@ -142,6 +143,27 @@ function Base.take!(buffer::CxxWrap.CxxWrapCore.SmartPointer{<:Buffer})
...
+#   push_back(!Matched::Union{CxxWrap.StdLib.StdVector{Any}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.StdVector{
+# Any}}}, ::Any)
+#    @ CxxWrap ~/.julia/packages/CxxWrap/aXNBY/src/CxxWrap.jl:624
+#   push_back(!Matched::Union{Ptr{Nothing}, CxxWrap.CxxWrapCore.CxxPtr{<:CxxWrap.StdLib.StdVector{Any}}}, ::Any)
+#    @ CxxWrap ~/.julia/packages/CxxWrap/aXNBY/src/CxxWrap.jl:624
+#   push_back(::Union{CxxWrap.StdLib.StdVector{CxxWrap.StdLib.SharedPtr{ray_core_worker_julia_jll.LocalMemoryBuffer
+# }}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.StdVector{CxxWrap.StdLib.SharedPtr{ray_core_worker_julia_jll.LocalMemoryBuffer}}}}, !Matched::Union{CxxWrap.CxxWrapCore.ConstCxxRef{<:CxxWrap.StdLib.SharedPtr{ray_core_worker_julia_jll.LocalMemoryBuffer}}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.SharedPtr{ray_core_worker_julia_jll.Local
+# MemoryBuffer}}, CxxWrap.CxxWrapCore.SmartPointer{T2} where T2<:ray_core_worker_julia_jll.LocalMemoryBuffer})
+#    @ ray_core_worker_julia_jll ~/.julia/packages/CxxWrap/aXNBY/src/CxxWrap.jl:624
+# ```
```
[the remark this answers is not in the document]
omus: I added these details to the issue you referenced.

deps/wrapper.cc
kleinschmidt: worth a shot

Ray.jl/src/runtime.jl
kleinschmidt: I'm kinda on the fence here; on the one hand, the variable names here could be used to communicate what the role these values play (I'd change `ray_objects` to `task_args` FWIW), and the function names then are free to describe what kind of values are being generated.  on the other hand, if we expect to _only_ use these for returns/args, then we shoudl just call them that.

Ray.jl/src/runtime.jl
```diff
@@ -131,9 +134,20 @@ function task_executor(ray_function, ray_objects)
...
     arg_string = join(string.("::", typeof.(args)), ", ")
     @info "Calling $func($arg_string)"
-    return func(args...)
+    result = func(args...)
+
+    # TODO: remove - useful for now for debugging
+    @info "Result: $result"
+
+    # TODO: support multiple return values
+    buffer_data = Vector{UInt8}(sprint(serialize, result))
```
kleinschmidt, line 146: is this better than printing to an IOBuffer and taking from it?

kleinschmidt (approved):
I think this is basically good.  some naming things as Curt raised but I'll leave it up to you to decide how to address my comments.  nothing blocking AFAIC

Ray.jl/src/runtime.jl
omus: Should be equivalent. Mainly this is just more terse.

deps/wrapper.cc
glennmoy: worked

deps/wrapper.cc
```diff
@@ -82,13 +84,17 @@ void initialize_coreworker_worker(
...
             bool is_streaming_generator) {
-            RAY_LOG(DEBUG) << "ray_core_worker_julia_jll: entered task_execuation_callback...";
-          // task_executor(ray_function, returns, args);
-          int pid = task_executor(ray_function, args);
-          std::string str = std::to_string(pid);
-          auto memory_buffer = std::make_shared<LocalMemoryBuffer>(reinterpret_cast<uint8_t *>(&str[0]), str.size(), true);
+
+          std::vector<std::shared_ptr<LocalMemoryBuffer>> return_vec;
+          auto args_ptr = static_cast<const void *>(&args);
+          task_executor(ray_function, &return_vec, args_ptr);
```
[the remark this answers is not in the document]
glennmoy: yeah I wonder if we supplied `&args` would it implicitly cast to a void pointer as well... I'll try it out

deps/wrapper.cc
```diff
@@ -100,6 +106,16 @@ void initialize_coreworker_worker(
...
 
+std::vector<std::shared_ptr<LocalMemoryBuffer>> * cast_to_buffer(void *ptr) {
+    auto buffer_ptr = static_cast<std::vector<std::shared_ptr<LocalMemoryBuffer>> *>(ptr);
+    return buffer_ptr;
+}
+
+std::vector<std::shared_ptr<RayObject>> cast_to_rayobject(void *ptr) {
+    auto rayobj_ptr = static_cast<std::vector<std::shared_ptr<RayObject>> *>(ptr);
+    return *rayobj_ptr;
+}
```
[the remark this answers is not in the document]
glennmoy: no...I should look into it though...the `buffer` being a ptr might be behind why we need to use a `push!` hack

Ray.jl/src/runtime.jl
glennmoy: Ok I take both your points and will rename the functions. will also rename `ray_object` to `task_args` while I'm at it.

deps/wrapper.cc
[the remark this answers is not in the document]
glennmoy: it does...which means we can rely on the implicit conversion if we like? I've added a comment calling out that this is happening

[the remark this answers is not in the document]
kleinschmidt: I actually kinda prefer the explicit cast so it's suuuper clear we're doing something fishy here.

deps/wrapper.cc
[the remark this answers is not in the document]
omus: I made the returns a pointer due to the issues we were having with mutability. It definitely may not be necessary though

omus (approved):
Looks good! I know there were a couple other follow up items to look into but it's definitely worth merging this now and making those changes as a follow up PR. Being limited to just returning `Int32`'s is definitely too restrictive.

omus:
@glennmoy I'm merging this now as having this available on `main` in it's current state it extremely useful. Feel free to make a follow up PR with additional changes you wanted to look into yet.

Issue #30, opened 2023-08-18
spawning second task fails silently


kleinschmidt:
Here's the behavior I've observed: from julia, I call `submit_task` once, which returns an object id.  then we can `get` that object which contains the function return value as expected.  when I call `submit_task` a second time, and try to `get` the object ID, it deadlocks julia (doesn't respond to Ctrl-C/SIGINT), presumably because it's waiting in the C++ code for a response from the object store.

The desired behavior is for teh second (and nth, more generally) task submission to behave like the first.

Okay, some more details/breadcrumbs:

The `raylet.err` file shows that it keeps trying to spawn workers to handle this task.  That's surprising to me given that (I'd have thought) the worker that executed the first task would be available again.  Here's an example of the log messages that seem relevant:

<details><summary>raylet.err: Initial worker startup</summary>

``` 
2023-08-18 10:51:54,670 DEBUG context.py:46 -- deserialize: {'command_prefix': [], 'env_vars': {'JULI
A_PROJECT': '/Users/dkleinschmidt/work/ray_core_worker_julia_jll.jl/Ray.jl'}, 'py_executable': '/User
s/dkleinschmidt/work/ray/venv/bin/python3.11', 'resources_dir': None, 'container': {}, 'java_jars': [
]}
2023-08-18 10:51:54,670 DEBUG context.py:51 -- Worker context env: {'JULIA_PROJECT': '/Users/dkleinsc
hmidt/work/ray_core_worker_julia_jll.jl/Ray.jl'}
2023-08-18 10:51:54,670 DEBUG context.py:99 -- Exec'ing worker with command: julia -e 'using Ray; sta
rt_worker()' -- /Users/dkleinschmidt/work/ray/python/ray/cpp/default_worker --ray_plasma_store_socket
_name=/tmp/ray/session_2023-08-18_10-51-24_326840_81975/sockets/plasma_store --ray_raylet_socket_name
=/tmp/ray/session_2023-08-18_10-51-24_326840_81975/sockets/raylet --ray_node_manager_port=52301 --ray
_address=127.0.0.1:6379 --ray_redis_password= --ray_session_dir=/tmp/ray/session_2023-08-18_10-51-24_
326840_81975 --ray_logs_dir=/tmp/ray/session_2023-08-18_10-51-24_326840_81975/logs --ray_node_ip_addr
ess=127.0.0.1 --runtime-env-hash=-916701441
[...]

PR #32, opened 2023-08-18
pass `--startup_token` through to julia workers

kleinschmidt:
Depends on https://github.com/beacon-biosignals/ray/pull/7.  Partially addresses #30 since it allows multiple workers to be created.  But doesn't fix the underlying issue of not re-using existing worker...

### DIFF
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -41,6 +41,7 @@ void initialize_coreworker_worker(
     std::string gcs_address,
     std::string node_ip_address,
     int node_manager_port,
+    int64_t startup_token,
     jlcxx::SafeCFunction julia_task_executor) {
     auto task_executor = jlcxx::make_function_pointer<int(
         RayFunction,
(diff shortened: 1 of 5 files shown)


kleinschmidt:
Will add an additional tests for multiple tasks!

deps/wrapper.cc
```diff
@@ -89,9 +91,14 @@ void initialize_coreworker_worker(
           (*returns)[0].second = std::make_shared<RayObject>(memory_buffer, nullptr, std::vector<rpc::ObjectReference>());
           return Status::OK();
         };
+    std::cerr << "Initializing julia worker coreworker" << std::endl;
     CoreWorkerProcess::Initialize(options);
 
+    std::cerr << "Starting julia worker task execution loop" << std::endl;
     CoreWorkerProcess::RunTaskExecutionLoop();
+
+    // doesn't seem to be reached...
```
glennmoy, line 100: do we need to include if it doesn't then?

glennmoy (approved):

deps/wrapper.cc
kleinschmidt: this was an act of desperation. I'll clean these up (or make them ray debug logs or something)

PR #34, opened 2023-08-21
Task executor now supports a single return

omus:


omus:
@glennmoy has incorporated these changes back into #31

PR #35, opened 2023-08-21
`task_executor` handles and reports errors

kleinschmidt:
This adds a `RayRemotException` a la `Distributed.RemoteException` that captures the task name and PID of the worker process, and the `CapturedException` caught on the worker, and returns that as the return from the task.  It also updates `Ray.get` to rethrow any `RayRemoteExceptions` that get got, a la `Base.fetch`.

depends on #31 for returning the exception ~~(just returns `Int32(0)` for now).~~

A few outstanding loose ends:

for some reason, ending in an error doesn't seem to complete the task, despite the fact that AFAICT we're doing the same thing that that the python worker/raylet stuff is doing.  will look into this more ~~tomorrow.~~ once MVP is complete

this does not handled nested tasks; ideally we should probably check that the caught exception is _not_ a `RayRemoteException` before doign additional wrapping, but IMO this is good enough for now.

---
  - https://app.asana.com/0/0/1205303690266617


kleinschmidt:
I thought that maybe the  workers not releasing after the task is complete was maybe due to not also setting `is_retryable_error` (in addition to the `application_error`, but no dice

Ray.jl/src/runtime.jl
```diff
@@ -122,29 +133,53 @@ function submit_task(f::Function, args::Tuple; runtime_env::RuntimeEnv=RuntimeEn
...
+        # so we use a cpp function whose only job is to assign the value to the
+        # pointer
+        err_msg = sprint(showerror, captured)
+        status = rayjll.report_error(application_error, err_msg, timestamp)
+        # XXX: we _can_ set _this_ return pointer here for some reason, and it
+        # was _harder_ to toss it back over the fence to the wrapper C++ code
+        is_retryable_error[] = rayjll.CxxBool(false)
+        @debug "push error status: $status"
+
+        result = RayRemoteException(getpid(), task_name, captured)
```
omus, line 181:
```suggestion
       RayRemoteException(getpid(), task_name, captured)
```

Ray.jl/test/task.jl
```diff
@@ -2,19 +2,23 @@
...
     # task with no args
     oid3 = submit_task(getpid, ())
-    result3 = deserialize(IOBuffer(take!(ray_core_worker_julia_jll.get(oid3))))
+    result3 = Ray.get(oid3)
     @test result3 isa Int32
     @test result3 != getpid()
+
+    # task that errors
+    oid4 = submit_task(error, ("AHHHHH",))
+    @test_throws Ray.RayRemoteException Ray.get(oid4)
```
omus, line 21: Should unpack this to confirm that an `ErrorException` is inside

src/wrappers/any.jl
```diff
@@ -172,9 +172,9 @@ function initialize_worker(raylet_socket, store_socket, ray_address, node_ip_add
     # obtained: Any
     # ```
     # Using `ConstCxxRef` doesn't seem supported (i.e. `const &`)
-    arg_types = (RayFunctionAllocated, Ptr{Cvoid}, Ptr{Cvoid})
-
-    # need to use `@eval` since `task_executor` is only defined at runtime
```
omus, line 177: Did you intend to remove this comment?

omus (approved):
Minor suggestions

Ray.jl/src/runtime.jl
```diff
@@ -122,29 +133,53 @@ function submit_task(f::Function, args::Tuple; runtime_env::RuntimeEnv=RuntimeEn
...
-    @info "Calling $func($arg_string)"
-    result = func(args...)
+        arg_string = join(string.("::", typeof.(args)), ", ")
+        @info "Calling $func($arg_string)"
+
+        result = func(args...)
+    catch e
+        # timestamp format to match python time.time()
+        # https://docs.python.org/3/library/time.html#time.time
+        timestamp = Dates.datetime2epochms(now()) / 1000
```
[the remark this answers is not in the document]
kleinschmidt: ahaha how did I miss `time()`??

src/wrappers/any.jl
kleinschmidt: I did not, good catch

Ray.jl/src/runtime.jl
```diff
@@ -122,29 +133,53 @@ function submit_task(f::Function, args::Tuple; runtime_env::RuntimeEnv=RuntimeEn
...
+            return deserialize(io)
+        end
 
-    arg_string = join(string.("::", typeof.(args)), ", ")
-    @info "Calling $func($arg_string)"
-    result = func(args...)
+        arg_string = join(string.("::", typeof.(args)), ", ")
+        @info "Calling $func($arg_string)"
+
+        result = func(args...)
```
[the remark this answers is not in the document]
kleinschmidt: yeah, I know it's not needed, but left it as a breadcrumb for future readers

omus:
Would be good to merge #47 before this PR

Ray.jl/src/object_store.jl
```diff
@@ -17,12 +17,17 @@ end
...
+if run in an `@async` task.
+
+If the task that generated the `ObjectID` failed with a Julia exception, the
+captured exception will be thrown on `get`.
 """
 function get(oid::rayjll.ObjectIDAllocated)
     io = IOBuffer(take!(rayjll.get(oid)))
-    return deserialize(io)
+    result = deserialize(io)
+    result isa RayRemoteException ? throw(result) : return result
```
omus, line 30: We'll probably want to be able to allow the caller to have the exception just be returned if they wish. I do think that this should be the default though and we can add that functionality in later

kleinschmidt: I'll add a TODO

Ray.jl/src/runtime.jl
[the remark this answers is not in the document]
kleinschmidt: yeah that's what I had originally actually, that's fine.

Issue #29, opened 2023-08-18
stochastic Deadline Exceeded error when running tests


glennmoy:
gets thrown from time to time, solution is to just rerun `] test`

```julia
(Ray) pkg> test
     Testing Ray
     ...
     ...
Precompiling project...
  1 dependency successfully precompiled in 1 seconds. 23 already precompiled.
     Testing Running tests...
[ Info: Starting local head node
2023-08-18 14:41:43,622 DEBUG node.py:1149 -- Process STDOUT and STDERR is being redirected to /tmp/ray/session_2023-08-18_14-41-43_620795_53189/logs.
2023-08-18 14:41:43,630 DEBUG node.py:605 -- Failed to send request to gcs, reconnecting. Error failed to connect to all addresses
2023-08-18 14:41:44,980 DEBUG node.py:1187 -- Process STDOUT and STDERR is being redirected to /tmp/ray/session_2023-08-18_14-41-43_620795_53189/logs.
2023-08-18 14:41:44,981 DEBUG services.py:1960 -- Determine to start the Plasma object store with 2.15 GB memory using /tmp.
calling delete on: 0x1356e61d0
function manager: Error During Test at /Users/glenn/.julia/dev/ray_core_worker_julia_jll/Ray.jl/test/function_manager.jl:23
  Test threw exception
  Expression: wait_for_function(fm, mfd, jobid; timeout_s = 1) == :timed_out
  Unknown: Deadline Exceeded
  Stacktrace:
   [1] Exists(arg1::Union{JuliaGcsClient, CxxWrap.CxxWrapCore.CxxRef{<:JuliaGcsClient}}, arg2::Union{AbstractString, CxxWrap.CxxWrapCore.ConstCxxRef{<:CxxWrap.StdLib.StdString}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.StdString}}, arg3::Union{AbstractString, CxxWrap.CxxWrapCore.ConstCxxRef{<:CxxWrap.StdLib.StdString}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.StdString}}, arg4::Integer)
     @ ray_core_worker_julia_jll ~/.julia/packages/CxxWrap/aXNBY/src/CxxWrap.jl:624
   [2] #9
     @ ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/function_manager.jl:76 [inlined]
[...]

PR #37, opened 2023-08-22
check for oversize functions before exporting

kleinschmidt:
This is based on the python checks:

thresholds defined here: https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/ray_constants.py#L120-L123

checks executed here: 

https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/utils.py#L725-L767

---
  - https://app.asana.com/0/0/1205303690266615

### DIFF
--- a/Ray.jl/src/function_manager.jl
+++ b/Ray.jl/src/function_manager.jl
@@ -42,7 +42,7 @@ function check_oversized_function(serialized, function_descriptor)
               "large ($(_mib_string(len))); FUNCTION_SIZE_ERROR_THRESHOLD=" *
               "$(_mib_string(FUNCTION_SIZE_ERROR_THRESHOLD)). " * check_msg
         throw(ArgumentError(msg))
-    elseif length(serialized) > FUNCTION_SIZE_WARN_THRESHOLD
+    elseif len > FUNCTION_SIZE_WARN_THRESHOLD
         msg = "The function $(rayjll.CallString(function_descriptor)) is very " *
               "large ($(_mib_string(len))). " * check_msg
         @warn msg
--- a/Ray.jl/src/function_manager.jl
+++ b/Ray.jl/src/function_manager.jl
@@ -28,8 +28,8 @@ using ray_core_worker_julia_jll: JuliaGcsClient, Exists, Put, Get,
                                  JuliaFunctionDescriptor, function_descriptor
 
 # https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/ray_constants.py#L122-L123
-const FUNCTION_SIZE_WARN_THRESHOLD = 10_000_000
-const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000
+const FUNCTION_SIZE_WARN_THRESHOLD = 10_000_000  # in bytes
+const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000  # in bytes
 
 _mib_string(len) = string(div(len, 1024 * 1024), " MiB")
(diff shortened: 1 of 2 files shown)


Ray.jl/src/function_manager.jl
```diff
@@ -27,6 +27,30 @@
 using ray_core_worker_julia_jll: JuliaGcsClient, Exists, Put, Get,
                                  JuliaFunctionDescriptor, function_descriptor
 
+const FUNCTION_SIZE_WARN_THRESHOLD = 10_000_000
+const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000
```
omus, line 31:
```suggestion
const FUNCTION_SIZE_WARN_THRESHOLD = 10_000_000  # in bytes
const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000  # in bytes
```

Ray.jl/src/function_manager.jl
```diff
@@ -27,6 +27,30 @@
...
+
+_mib_string(len) = string(div(len, 1024 * 1024), " MiB")
+
+function check_oversized_function(serialized, function_descriptor)
+    len = length(serialized)
+    check_msg = "Check that its definition is not implicitly capturing a large " *
+                "array or other object in scope. Tip: use `Ray.put()` to put large " *
+                "objects in the Ray object store."
```
omus, line 39: Are these messages based upon a message from Python Ray? May be good to include a reference here

omus: Should be made into a global constant

Ray.jl/test/function_manager.jl
```diff
@@ -48,6 +48,24 @@
     fc2 = import_function!(fm, fcd, jobid)
     @test fc.(1:10) == fc2.(1:10) != f.(1:10)
 
+    @testset "warn/error for large functions" begin
+        # for some reason, using `let` to introduce local scope does not work
+        # during test execution to generate a closure, even though it works
+        # locally.  so we use a factory intsead:
```
omus, line 54:
```suggestion
        # locally, so we use a factory instead:
```

omus (approved):

Ray.jl/test/function_manager.jl
```diff
@@ -48,6 +48,24 @@
...
+        function bigfactory(size)
+            x = rand(UInt8, size)
+            return f(y) = y * sum(x)
```
glennmoy, line 57: might be better to locally define a new `f` since it's not obvious when it's declared that another unit test depends on it?

Ray.jl/test/function_manager.jl
```diff
@@ -13,7 +13,7 @@
     export_function!(fm, f, jobid)
 
     fd = function_descriptor(f)
-    @test wait_for_function(fm, fd, jobid) == :ok
+    @test wait_for_function(fm, fd, jobid; timeout_s=10) == :ok
```
glennmoy, line 16: ?

glennmoy (approved):

Ray.jl/test/function_manager.jl
kleinschmidt: it ends up being an anonymous function anyway (due to the local scope)

Ray.jl/test/function_manager.jl
kleinschmidt: this is whats causing #29

kleinschmidt: (the timeout being too short)

PR #38, opened 2023-08-23
Support package loading on workers via `RuntimeEnv`

omus:
Supports [package loading](https://docs.julialang.org/en/v1/manual/code-loading/#Definitions) for task workers by allowing users to specify a runtime environment which specifies the package loading statements to use.

Ray for Python handles this via their custom serialization code `ray.cloudpickle` which handles serialization of global references. We may want to handle things similarly in Julia with an alternative serializer but doing so would be a large undertaking which is unnecessary at this point.

### DIFF
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -248,7 +248,7 @@ function start_worker(args=ARGS)
     _init_global_function_manager(parsed_args["address"])
 
     # Load top-level package loading statements (e.g. `import X` or `using X`) to ensure
-    ## tasks have access to dependencies.
+    # tasks have access to dependencies.
     if haskey(ENV, "JULIA_RAY_PACKAGE_IMPORTS")
         io = IOBuffer(ENV["JULIA_RAY_PACKAGE_IMPORTS"])
         pkg_imports = deserialize(Base64DecodePipe(io))
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -1,3 +1,12 @@
+const JOB_RUNTIME_ENV = Ref{RuntimeEnv}()
+
+macro ray_import(ex)
+    isassigned(JOB_RUNTIME_ENV) && error("`@ray_import` can only be called once")
+    runtime_env = RuntimeEnv(; package_imports=ex)
+    JOB_RUNTIME_ENV[] = runtime_env
+    return esc(ex)
+end
+
 """
     const GLOBAL_STATE_ACCESSOR::Ref{rayjll.GlobalStateAccessor}
 
@@ -8,7 +17,7 @@ This is set during `init` and used there to get the Job ID for the driver.
 """
 const GLOBAL_STATE_ACCESSOR = Ref{rayjll.GlobalStateAccessor}()
 
-function init(runtime_env::RuntimeEnv=RuntimeEnv())
+function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing)
     # XXX: this is at best EXREMELY IMPERFECT check.  we should do something
     # more like what hte python Worker class does, getting node ID at
     # initialization and using that as a proxy for whether it's connected or not
@@ -17,6 +26,16 @@ function init(runtime_env::RuntimeEnv=RuntimeEnv())
     if isassigned(FUNCTION_MANAGER)
         @warn "Ray already initialized, skipping..."
         return nothing
+    end
+
+    if isnothing(runtime_env)
+        # Set default for `JOB_RUNTIME_ENV` when `Ray.init` is called before `@ray_import`.
+        # This ensures a call to `@ray_import` after `Ray.init` will fail.
+        if !isassigned(JOB_RUNTIME_ENV)
+            JOB_RUNTIME_ENV[] = RuntimeEnv()
+        end
+
+        runtime_env = JOB_RUNTIME_ENV[]
     end
 
     # TODO: use something like the java config bootstrap address (?) to get this
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -1,12 +1,23 @@
 const JOB_RUNTIME_ENV = Ref{RuntimeEnv}()
 
 macro ray_import(ex)
+    Ray = gensym(:Ray)
+    result = quote
+        import Ray as $Ray
+        $Ray._ray_import($Ray.RuntimeEnv(; package_imports=$(QuoteNode(ex))))
+        $ex
+    end
+
+    return esc(result)
+end
+
+function _ray_import(runtime_env::RuntimeEnv)
     if isassigned(JOB_RUNTIME_ENV)
         error("`@ray_import` must be used before `Ray.init` and can only be called once")
     end
-    runtime_env = RuntimeEnv(; package_imports=ex)
+
     JOB_RUNTIME_ENV[] = runtime_env
-    return esc(ex)
+    return nothing
 end
 
 struct RayRemoteException <: Exception
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -1,7 +1,9 @@
 const JOB_RUNTIME_ENV = Ref{RuntimeEnv}()
 
 macro ray_import(ex)
-    isassigned(JOB_RUNTIME_ENV) && error("`@ray_import` can only be called once")
+    if isassigned(JOB_RUNTIME_ENV)
+        error("`@ray_import` must be used before `Ray.init` and can only be called once")
+    end
     runtime_env = RuntimeEnv(; package_imports=ex)
     JOB_RUNTIME_ENV[] = runtime_env
     return esc(ex)
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -9,6 +9,17 @@ macro ray_import(ex)
     return esc(ex)
 end
 
+struct RayRemoteException <: Exception
+    pid::Int
+    task_name::String
+    captured::CapturedException
+end
+
+function Base.showerror(io::IO, re::RayRemoteException)
+    print(io, "on Ray task \"$(re.task_name)\" with PID $(re.pid): ")
+    showerror(io, re.captured)
+end
+
 """
     const GLOBAL_STATE_ACCESSOR::Ref{rayjll.GlobalStateAccessor}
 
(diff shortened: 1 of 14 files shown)


Ray.jl/test/task.jl
```diff
@@ -17,3 +17,21 @@
...
+@testset "Task RuntimeEnv" begin
+    @testset "package_imports" begin
+        f = x -> Int32(length(string(nameof(Test))))
+        runtime_env = Ray.RuntimeEnv(; package_imports=:(using Test))
+        oid = submit_task(f, (1,); runtime_env)
+        result = String(take!(ray_core_worker_julia_jll.get(oid)))
+        @test parse(Int, result) == 4
+
+        # The spawned worker will fail with "ERROR: UndefVarError: `Test` not defined". For
+        # now since we have worker exception handling we'll detect this by attempting to
```
kleinschmidt, line 30:
```suggestion
        # now since we don't have worker exception handling we'll detect this by attempting to
```

?

kleinschmidt (changes requested):
Given discussion with @jrevels about constraints on global state between workers/driver I'm now not sure this is a good idea.  especially since it basically allows _arbitrary_ code to be executed on worker startup.  I think we should keep this in our back pocket for now _unless_ we find that we _really_ need this level of flexibilty for MVP.

omus:
> Given discussion with @jrevels about constraints on global state between workers/driver I'm now not sure this is a good idea. especially since it basically allows _arbitrary_ code to be executed on worker startup. I think we should keep this in our back pocket for now _unless_ we find that we _really_ need this level of flexibilty for MVP.

I'll point out that a Ray task is to allow code execution on worker startup. I agree we don't want people abusing this feature but we can easily limit the expressions allowed here to avoid any real abuse.

If we want to go in a completely different direction then I'm not sure what our approach should be. To start with lets recap some facts:

- Julia package loading statements are only allowed at the top-level.
- Without package loading statements Julia Ray tasks will be limited to being able to only calling `Base` and `Ray` functions (the later is due to us starting Julia with `using Ray; start_worker()`)
- Julia's serialization does not allow you to deserialize a function into a specified module. This limits our ability to isolate Julia Ray tasks to run with an isolated module.
- Python Ray tasks serialize their global references allowing them to import dependencies during deserialization.

For us to support package loading we can:

1. Include package imports in the function store:
    a. Utilize a custom serializer similar to `ray.cloudpickle` so that a single `deserialize` call also handles package loading
    b. Use sidecar data such that package imports and the serialized function are stored together but are deserialized separately. Would require users to specify the imports like in this PR
    c. Use `@eval` within the function definition to evalulate things at the top-level (e.g. `f = () -> (@eval using Dates; now())`)
2. Include package imports in the runtime environment (this PR). IIRC runtime environments are hashed and only tasks using the same runtime environment hash can be scheduled together. As runtime environments can be specified per-job or per-task by specifying different package imports would result in two tasks that could be run on one worker being scheduled on different ones. I was planning on working around this problem by only using per-job imports to start with.

kleinschmidt:
Definitely agree that _some_ mechanism for loading dependencies on the workers is needed; the question is how much leeway we should be giving users to determine _what_ gets loaded on workers.  I agree with @jrevels that allowing _any differences_ between the driver and worker environments is a huge footgun and we should put strong constraints on that.  For an MVP it might be fine to make those constraints informal, and just say "yeah if there's any difference it's undefined behavior" but what I don't wanna be doing is introducing a backdoor or wiggle room that will be hard to walk back later when we're in teh process of handing things off to actual users.

One approach we could take is to fix the loaded code at `Ray.init()` time, using `Base.loaded_modules` (or one of the variants thereof) to determine which of teh project packages have been loaded, and send that list over as part of the runtime environment.  The tradeoff there is that we take away teh ability to differentiate between things like import/using, but that is IMO a fine tradeoff to make (the import/using thing is really more of a concern when you're developing a package where it matters what gets brought into a particular namespace).

Another approach would be to require that every Ray job have a "context package" which is the _only_ thing that gets loaded.  That's similar to one of teh workflows we've used commonly at Beacon (where a large project that needs to run in a distributed setting has a single entrypoint package that loads all dependencies and defines the work functions, and provides a "provision workers" function that does a `workers = addprocs(...); remotecall_eval(workers, :(using MyProjectPackage))`).  Then we can auto-detect the active project and send that as part of the job's runtime environment.

omus:
> One approach we could take is to fix the loaded code at `Ray.init()` time, using `Base.loaded_modules` (or one of the variants thereof) to determine which of teh project packages have been loaded, and send that list over as part of the runtime environment. The tradeoff there is that we take away teh ability to differentiate between things like import/using, but that is IMO a fine tradeoff to make (the import/using thing is really more of a concern when you're developing a package where it matters what gets brought into a particular namespace).

When using `Base.loaded_modules` we'd be limited to either having to fully qualify all calls (`Dates.now` vs. `now`) for any function we wish to run as a Ray task. One definite gotcha with this is that function that run fine locally would need to be modified to run on a worker. Alternatively, if we used `using` then we'd be less limited by this but ambiguous exports or explicitly imported functions are still problematic.

My point with bringing this up is that even with the proposed approach we're still in a scenario where the driver and worker will have different environments.

> Another approach would be to require that every Ray job have a "context package" which is the _only_ thing that gets loaded. That's similar to one of teh workflows we've used commonly at Beacon (where a large project that needs to run in a distributed setting has a single entrypoint package that loads all dependencies and defines the work functions, and provides a "provision workers" function that does a `workers = addprocs(...); remotecall_eval(workers, :(using MyProjectPackage))`). Then we can auto-detect the active project and send that as part of the job's runtime environment.

This approach definitely forces users to work in a very specific way. I believe this will hamper interactive Ray development which may be done directly in the REPL or via a single included file.

> I agree with @jrevels that allowing _any differences_ between the driver and worker environments is a huge footgun and we should put strong constraints on that.

I also agree with this. However, we may want to consider avoiding loading unnecessary packages on a worker to be a performance optimization. With this in mind what we want is that any code executed on a worker should behave identically to how it would be run on the driver. Complications due to extended functions or evaluated code may mean we have to load all the same packages but we should keep in mind this distinction.

I'll see what other approaches I can come up with.

kleinschmidt:
> When using Base.loaded_modules we'd be limited to either having to fully qualify all calls (Dates.now vs. now) for any function we wish to run as a Ray task. One definite gotcha with this is that function that run fine locally would need to be modified to run on a worker. Alternatively, if we used using then we'd be less limited by this but ambiguous exports or explicitly imported functions are still problematic.

Ah yeah, I see that now:

```julia
julia> using Distributed

julia> addprocs(1)
1-element Vector{Int64}:
 2

julia> using Dates

julia> now()
2023-08-23T16:34:01.728

julia> f = () -> now()
#3 (generic function with 1 method)

julia> @everywhere using Serialization

julia> fbytes = sprint(serialize, f);

julia> @everywhere 2 import Dates

julia> remotecall_fetch(only(workers()), fbytes) do fbytes
           io = IOBuffer(fbytes)
           f = deserialize(io)
           return invokelatest(f)
       end
ERROR: On worker 2:
UndefVarError: `now` not defined

julia> @everywhere 2 using Dates

julia> remotecall_fetch(only(workers()), fbytes) do fbytes
           io = IOBuffer(fbytes)
           f = deserialize(io)
           return invokelatest(f)
       end
2023-08-23T16:37:20.130
```

the issue is that the serialized function does not capture the module where global refs are actually defined.

kleinschmidt:
another possibility that came up on the julia slack is to use the callbacks that revise/distributed.jl hook into: https://julialang.slack.com/archives/C67EFTEF3/p1692818861689979

omus:
This is now functional:
```julia
using Ray

@ray_import begin
    using Dates
end

Ray.init()

f = () -> now()
ref = submit_task(f, ())
Ray.get(ref)
```

I've currently left in the ability to pass in a `RuntimeEnv` to `Ray.init` (per-job) or via `submit_task` (per-task). I can definitely remove these so to avoid any user confusion but my personal preference is to at least include them in one commit for reference. Additionally, the per-task runtime environments are definitely more test friendly. I've included per-job tests as the code path is different enough but I've had to spawn temporary Julia processes to work around our "`Ray.init` can only be called once" logic.

Ray.jl/src/runtime.jl
```diff
@@ -30,6 +41,16 @@ function init()
         return nothing
     end
 
+    if isnothing(runtime_env)
+        # Set default for `JOB_RUNTIME_ENV` when `Ray.init` is called before `@ray_import`.
+        # This ensures a call to `@ray_import` after `Ray.init` will fail.
```
kleinschmidt, line 46: this isn't really accurate given that the `@ray_import` code that sets this ref is run at parse time (it's in the macro body, not the returned expression) not run time, right?

deps/wrapper.cc
```diff
@@ -313,6 +320,31 @@ Status report_error(std::string *application_error,
     return worker.PushError(jobid, "task", err_msg, timestamp);
 }
 
+// Serialize a JobConfig protobuf message into its serialized string equivalent by constructing it
+// via it's JSON representation of the message. Most likely there is a better way to construct
+// the protobuf message but this is works for now.
```
kleinschmidt, line 325: yeah, probably a better way to do this but given that we're not setting this per-task, I think it's fine if it's a bit inefficient and worth the tradeoff in not getting too rabbit holed here

Ray.jl/test/utils.jl
```diff
@@ -21,3 +21,58 @@ function setup_core_worker(body)
         rayjll.shutdown_driver()
     end
 end
+
+# Useful in running tests which require us to re-run `Ray.init` which currently can only be
+# called once as it is infeasible to reset global `Ref`'s using `isassigned` conditionals.
+macro process_eval(ex)
+    expr = QuoteNode(ex)
+    return esc(:(process_eval($expr)))
+end
```
kleinschmidt, line 30: could mock `init` too?

Ray.jl/src/runtime_env.jl
```diff
@@ -1,15 +1,80 @@
...
+    end
 end
 
-RuntimeEnv(; project=project_dir()) = RuntimeEnv(project)
+function RuntimeEnv(; project=project_dir(), package_imports=Expr(:block))
+    return RuntimeEnv(project, package_imports)
+end
 
 function json_dict(runtime_env::RuntimeEnv)
+    env_vars = Dict("JULIA_PROJECT" => runtime_env.project)
```
kleinschmidt, line 15: probably good to at least add a TODO to support user-provided `env_vars` in the runtime env too.

kleinschmidt (commented):
only one major thing to sort out here.

Ray.jl/src/runtime.jl
```diff
@@ -1,3 +1,14 @@
+const JOB_RUNTIME_ENV = Ref{RuntimeEnv}()
+
+macro ray_import(ex)
+    if isassigned(JOB_RUNTIME_ENV)
+        error("`@ray_import` must be used before `Ray.init` and can only be called once")
+    end
+    runtime_env = RuntimeEnv(; package_imports=ex)
+    JOB_RUNTIME_ENV[] = runtime_env
```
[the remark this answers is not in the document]
omus: > Hm, can you explain a bit more why you do this stuff at load/macroexpand time and not at run time?

Originally I was thinking that calling `RuntimeEnv` at load time was just more performance but thinking about it more it'll definitely get messed up with PackageCompiler or even with `.ji` files if you change your project dir. I'll update this.

Ray.jl/src/runtime.jl
omus: You are correct the comment isn't accurate. When running in the REPL however the macro can be expanded after `Ray.init`

Ray.jl/src/runtime_env.jl
```diff
@@ -1,15 +1,80 @@
...
+
+    # Avoid including package imports if the expression is an empty block
+    imports = runtime_env.package_imports
+    if imports.head !== :block || !isempty(imports.args)
+        env_vars["JULIA_RAY_PACKAGE_IMPORTS"] = base64encode(serialize, imports)
```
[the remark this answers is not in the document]
omus: Probably is overkill. As far as I'm aware there isn't a size limit here

Ray.jl/test/utils.jl
omus: I considered using Mocking for this but realized quickly for the tests which validate we don't accidentally load unnecessary packages by default would be extremely hard to get right and went with this approach instead.

kleinschmidt: ah yeah taht's a good point

Ray.jl/src/runtime.jl
omus: Comment is now correct given that we set `JOB_RUNTIME_ENV` at runtime

Ray.jl/src/runtime_env.jl
```diff
@@ -1,15 +1,80 @@
...
+    return rayjll.serialize_job_config_json(job_config_json)
+end
+
+function process_import_statements(ex::Expr)
+    if ex.head === :using || ex.head === :import
+        return ex
+    elseif ex.head === :block
+        imports = Expr(:block)
+        for arg in ex.args
+            arg isa LineNumberNode && continue
```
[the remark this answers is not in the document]
omus: I'll add a comment about this as it's not obvious

kleinschmidt (approved):
I'm okay with these changes as they are but CI is failing on 1.8 so we need to get that sorted out.  Looks like it needs `using Serialization` in some of the tests, and the updated test for missing imports isn't working as intended on 1.8

once CI is happy go ahead and merge!

omus:
> I'm okay with these changes as they are but CI is failing on 1.8 so we need to get that sorted out. Looks like it needs `using Serialization` in some of the tests, and the updated test for missing imports isn't working as intended on 1.8
> 
> once CI is happy go ahead and merge!

Great! I'm looking into the CI issue. I'll mention I didn't see the failure on my local machine.

Ray.jl/test/task.jl
```diff
@@ -36,4 +36,116 @@ end
...
+    @testset "package_imports" begin
+        f = () -> nameof(Test)
+        runtime_env = Ray.RuntimeEnv(; package_imports=:(using Test))
+        result = Ray.get(submit_task(f, (); runtime_env))
+        @test result == :Test
+
+        # The spawned worker will fail with "ERROR: UndefVarError: `Test` not defined". We
+        # can detect this failure attempting to fetch the task result.
+        ref = submit_task(f, ())
+        msg = if VERSION <= v"1.8"
```
omus, line 49: Made a very easy to make mistake:
```julia
julia> v"1.8.1" <= v"1.8"
false
```

PR #39, opened 2023-08-23
Run Ray.jl CI workflow on draft PRs

omus:
This is slowing us down more than it's helping at the moment. I also removed some unnecessary workflow changes.


omus:
Starting this PR in draft mode to verify that the Ray.jl CI workflow runs in draft mode (it does)

kleinschmidt (approved):

PR #40, opened 2023-08-23
tasks can run no-arg functions

glennmoy:
---
  - https://app.asana.com/0/0/1205338201542037

### DIFF
--- a/Ray.jl/test/task.jl
+++ b/Ray.jl/test/task.jl
@@ -15,5 +15,5 @@
     oid3 = submit_task(getpid)
     result3 = String(take!(ray_core_worker_julia_jll.get(oid3)))
     @test all(isdigit, result2)
-    @test parse(Int32 result2) != getpid()
+    @test parse(Int32, result2) != getpid()
 end
--- a/Ray.jl/test/task.jl
+++ b/Ray.jl/test/task.jl
@@ -10,4 +10,10 @@
     result2 = String(take!(ray_core_worker_julia_jll.get(oid2)))
     @test all(isdigit, result2)
     @test parse(Int, result2) == 6
+
+    # task with no args
+    oid3 = submit_task(getpid)
+    result3 = String(take!(ray_core_worker_julia_jll.get(oid3)))
+    @test all(isdigit, result2)
+    @test parse(Int32 result2) != getpid()
 end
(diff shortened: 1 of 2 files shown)


src/wrappers/any.jl
```diff
@@ -144,7 +144,11 @@ end
 
 # XXX: Need to convert julia vectors to StdVector. This function helps us avoid having
 # CxxWrap as a direct dependency in Ray.jl
-_submit_task(dir, fd, oids::AbstractVector) = _submit_task(dir, fd, StdVector(oids))
+function _submit_task(dir, fd, oids::AbstractVector)
+    # https://github.com/JuliaInterop/CxxWrap.jl/issues/367
+    args = isempty(oids) ? StdVector{ObjectID}() : StdVector(oids)
```
glennmoy, line 149: ideally `StdVector(ObjectID[])` would be equivalent to `StdVector{ObjectID}()`

Ray.jl/test/task.jl
```diff
@@ -10,4 +10,10 @@
     result2 = String(take!(ray_core_worker_julia_jll.get(oid2)))
     @test all(isdigit, result2)
     @test parse(Int, result2) == 6
+
+    # task with no args
+    oid3 = submit_task(getpid)
+    result3 = String(take!(ray_core_worker_julia_jll.get(oid3)))
+    @test all(isdigit, result2)
+    @test parse(Int32 result2) != getpid()
```
glennmoy, line 18:
```suggestion
    @test parse(Int32, result2) != getpid()
```

src/wrappers/any.jl
omus: I was going to advocate that we just pass in a `ObjectID[]` given your statement above but what you intended to convey is that these `StdVector(ObjectID[])` and `StdVector{ObjectID}()` are not the same even though they are equal:
```julia
julia> StdVector(ObjectID[])
0-element CxxWrap.StdLib.StdVectorAllocated{Any}

julia> StdVector{ObjectID}()
0-element CxxWrap.StdLib.StdVectorAllocated{ObjectID}

julia> StdVector(ObjectID[]) == StdVector{ObjectID}()
true
```

src/wrappers/any.jl
```diff
@@ -144,7 +144,11 @@ end
...
+    _submit_task(dir, fd, args)
```
omus, line 150:
```suggestion
    return _submit_task(dir, fd, args)
```

omus (approved):

src/wrappers/any.jl
glennmoy: yes! I didn't actually check equality though (assumed it would fail). But since they have different parameter types the first one doesn't get dispatched to the Cxx `_submit_task`.

PR #42, opened 2023-08-23
re-use workers for multiple tasks

kleinschmidt:
the issue was that we were not setting `runtime_env_hash` correctly in the core worker options.

~~the test needs work: currently, it'll spin up more workers to handle the extra load.  informally when I tested htis locally, running the test twice, it does re-use 8 workers (number of CPUs on my macbook air).~~

---
  - https://app.asana.com/0/0/1205303690266614

### DIFF
--- a/Ray.jl/test/task.jl
+++ b/Ray.jl/test/task.jl
@@ -149,3 +149,21 @@ end
         end
     end
 end
+
+@testset "many tasks" begin
+    oids = map(1:20) do i
+        s = 20
+        submit_task(s, i) do s, i
+            println(stderr, "task $i running on PID $(getpid()) sleeping $s...")
+            flush(stderr)
+            sleep(s)
+            println("task $i running on PID $(getpid()) slept $s")
+            flush(stderr)
+            # need an Int32 return value for now
+            return Int32(i)
+        end
+    end
+
+    _get(x) = String(take!(ray_core_worker_julia_jll.get(x)))
+    @time map(_get, oids)
+end
--- a/Ray.jl/test/task.jl
+++ b/Ray.jl/test/task.jl
@@ -153,7 +153,7 @@ end
 @testset "many tasks" begin
     oids = map(1:20) do i
         s = 20
-        submit_task(s, i) do s, i
+        submit_task((s, i)) do s, i
             pid = getpid()
             println(stderr, "task $i running on PID $pid sleeping $s...")
             flush(stderr)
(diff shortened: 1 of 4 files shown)


deps/wrapper.cc
```diff
@@ -79,6 +80,7 @@ void initialize_worker(
     options.raylet_ip_address = node_ip_address;
     options.metrics_agent_port = -1;
     options.startup_token = startup_token;
+    options.runtime_env_hash = runtime_env_hash;
```
glennmoy, line 83: huh I actually came across this last week and tried harcoding `runtime_env_hash` but found it still didn't work

https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/raylet/worker.h#L274-L277

glennmoy (approved):
LGTM pending an appropriate test

deps/wrapper.cc
glennmoy: ah this actually corroborates what you saw in the logs about it not being properly propagated! `runtime_env_hash: 1462190447 cached runtime env hash: 0`

kleinschmidt: ah you were on teh right track long before I was!

Ray.jl/test/task.jl
```diff
@@ -151,19 +151,10 @@ end
...
-    end
-
-    @time pids = Ray.get.(oids)
-    @test !allunique(pids)
+    n_tasks = Sys.CPU_THREADS * 2
+    # warm up worker cache pool
+    pids = Ray.get.([submit_task(getpid, ()) for _ in 1:n_tasks])
+    # run more tasks which should re-use the cpu cache
+    pids2 = Ray.get.([submit_task(getpid, ()) for _ in 1:n_tasks])
+    @test !isempty(intersect(pids, pids2))
```
glennmoy, line 159: test is effectively saying "at least one process should get re-used" which is probably good enough to show it works

glennmoy (approved):

Ray.jl/test/task.jl
```diff
@@ -149,3 +149,12 @@ end
         end
     end
 end
+
+@testset "many tasks" begin
+    n_tasks = Sys.CPU_THREADS * 2
```
kleinschmidt, line 154: we can  maybe get away with smaller number of tasks but want to make sure we saturate teh cache of workers

PR #43, opened 2023-08-24
Implement Ray.jl put/get

glennmoy:
---
  - https://app.asana.com/0/0/1205319616948328

### DIFF
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -1,0 +1,14 @@
+function put(data)
+    bytes = Vector{UInt8}()
+    io = IOBuffer(bytes; write=true)
+    serialize(io, data)
+    buffer_ptr = Ptr{Nothing}(pointer(bytes))
+    buffer_size = sizeof(bytes)
+    buffer = rayjll.LocalMemoryBuffer(buffer_ptr, buffer_size, true)
+    return rayjll.put(buffer)
+end
+
+function get(oid::rayjll.ObjectIDAllocated)
+    io = IOBuffer(take!(rayjll.get(oid)))
+    return deserialize(io)
+end
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -29,4 +29,4 @@ end
 # TODO: support any collection? Python ray only supports lists, not tuples, don't see why not
 get(oids::AbstractVector) = map(get, oids)
 
-# TOOD: fallback get(x) = x? Python ray errors but no good reason not to do this
+get(x) = x
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -1,8 +1,8 @@
 """
     Ray.put(data) -> ObjectIDAllocated
 
-Stores `data` in the object store and returns an object reference assigned to the value that
-can used to retrieve the data with [`Ray.get`](@ref).
+Store `data` in the object store. Returns an object reference which can used to retrieve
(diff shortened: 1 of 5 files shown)


Ray.jl/src/object_store.jl
```diff
@@ -0,0 +1,30 @@
...
+Retrieves the data associated with the (collection of) `object_id`(s) from the object store.
+This method is blocking until the data is available in the local object store.
+"""
+function get(oid::rayjll.ObjectIDAllocated)
+    io = IOBuffer(take!(rayjll.get(oid)))
+    return deserialize(io)
+end
+
+get(oids::Union{AbstractVector, Tuple}) = map(get, oids)
+get(x) = x
```
glennmoy, line 30: I'll call out that ray Python only supposes _lists_ of object refs, not tuples, and doesn't have a fallback method either.

I didn't see a reason not to include these, other than to be consistent with what Python doesn't support.
Supporting common collections of object refs and not erroring over non-objectids seems a no brainer?

Ray.jl/test/object_store.jl
```diff
@@ -0,0 +1,26 @@
+@testset "object_store.jl" begin
+
+    @testset "Put/Get roundtrip for $(typeof(x))" for x in (
+        1, 1.23, "hello", (1, 2, 3), [1, 2, 3],
```
glennmoy, line 4: any other types worth checking? might already be overdoing it

Ray.jl/src/object_store.jl
```diff
@@ -0,0 +1,30 @@
+"""
+    Ray.put(data) -> ObjectIDAllocated
```
glennmoy, line 2: adding (minimal) docstrings seeing as these are now our high-level user-facing functions

Ray.jl/src/object_store.jl
omus: I'll mention I'm for having the `get(x) = x` fallback method. Having this will be important for writing generic Julia functions which may receive a `ObjectID` or may receive the value directly depending on whether it's called locally or as a remote Ray task.

glennmoy: makes sense!

Ray.jl/src/object_store.jl
```diff
@@ -0,0 +1,29 @@
...
+    serialize(io, data)
+    buffer_ptr = Ptr{Nothing}(pointer(bytes))
+    buffer_size = sizeof(bytes)
+    buffer = rayjll.LocalMemoryBuffer(buffer_ptr, buffer_size, true)
+    return rayjll.put(buffer)
+end
+
+"""
+    Ray.get(object_id::ObjectIDAllocated)
+    Ray.get(object_ids::Union{AbstractVector, Tuple})
```
omus, line 19:
```suggestion
```

Ray.jl/src/runtime.jl
```diff
@@ -99,8 +99,6 @@ end
 
 initialize_coreworker_driver(args...) = rayjll.initialize_coreworker_driver(args...)
 
-project_dir() = dirname(Pkg.project().path)
```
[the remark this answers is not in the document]
omus: Lets merge https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/pull/44 first so we have a better record

omus (approved):
Thanks. If you can approve and merge https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/pull/44 that would be great!

PR #45, opened 2023-08-24
Rename initialize/shutdown functions

omus:
Quick cleanup pass:

- Rename `initialize_coreworker_driver` to `initialize_driver`
- Rename `initialize_coreworker_worker` to `initialize_worker`
- Rename `shutdown_coreworker` to `shutdown_driver`
- Rename `ray_core_worker_julia_jll.start_worker` to `ray_core_worker_julia_jll.initialize_worker`

### DIFF
--- a/Ray.jl/src/Ray.jl
+++ b/Ray.jl/src/Ray.jl
@@ -13,11 +13,9 @@ using LoggingExtras
 using Pkg
 using Serialization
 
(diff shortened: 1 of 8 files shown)


glennmoy (approved):
just confirming we don't need to update the `start_worker` call in `ray/` since I don't think we touched that function here?

PR #47, opened 2023-08-25
Use high-level get in tests

omus:
Follow up to #43 to have the task tests use the high-level `Ray.get`/`Ray.put`. This change allowed us to remove some loaded packages in the tests.

Additionally, I removed the unused test dependency on Distributed and Pkg which were introduced in #19.

### DIFF
--- a/Ray.jl/test/runtests.jl
+++ b/Ray.jl/test/runtests.jl
@@ -1,10 +1,7 @@
 using Aqua
-using Distributed
 using Ray
-using Serialization
 using Test
 
-using ray_core_worker_julia_jll
 import ray_core_worker_julia_jll as rayjll
 
 include("setup.jl")
@@ -20,8 +17,8 @@ include("utils.jl")
     setup_ray_head_node() do
         include("function_manager.jl")
         setup_core_worker() do
-            include("task.jl")
             include("object_store.jl")
+            include("task.jl")
         end
     end
 end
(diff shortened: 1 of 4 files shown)


omus:
Some of the work here was integrated into #35

Ray.jl/test/runtests.jl
```diff
@@ -20,8 +17,8 @@ include("utils.jl")
     setup_ray_head_node() do
         include("function_manager.jl")
         setup_core_worker() do
-            include("task.jl")
             include("object_store.jl")
+            include("task.jl")
```
kleinschmidt, line 21: why swap the order of these?

Ray.jl/test/task.jl
```diff
@@ -1,30 +1,27 @@
 @testset "Submit task" begin
-    oid1 = submit_task(length, ("hello",))
-    oid2 = submit_task(max, (0x00, 0xff))
```
kleinschmidt, line 3: making these concurrent was intentional, that way we at least stress the path where there are >1 task in flight.  not the biggest deal though adn we shoudl probably be testing that separately/explicitly.

kleinschmidt (approved):

Ray.jl/test/runtests.jl
omus: Because `task.jl` now depends on `Ray.get`. Makes more sense to test `object_store.jl` first

Ray.jl/test/task.jl
omus: Multiple tasks in flight should definitely be tested separately. I noticed the way we laid out the tests before it was easy to forget to update the suffix on `oid` or `result` and accidentally get a test pass. This just avoids those kinds of copy/paste issues

omus: https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/issues/50

PR #49, opened 2023-08-28
Support executing tasks with keywords

omus:
- [x] Requires #47.


Ray.jl/src/runtime.jl
```diff
@@ -181,16 +175,21 @@ function task_executor(ray_function, returns_ptr, task_args_ptr, task_name,
                                 rayjll.unwrap_function_descriptor(fd),
                                 get_current_job_id())
         # TODO: write generic Ray.put and Ray.get functions and abstract over this buffer stuff
-        args = map(task_args) do arg
+        flattened = map(task_args) do arg
```
glennmoy, line 178: could we map over `Ray.get` here?

Ray.jl/src/remote_function.jl
```diff
@@ -0,0 +1,34 @@
+# Underscores are not allowed to be used as Julia function keywords
+# https://docs.julialang.org/en/v1/manual/variables/#man-allowed-variable-names
+const ARG_KEY = :_
+
+function flatten_args(args, kwargs)
```
glennmoy, line 5: maybe write a cursory unit test for this?

Ray.jl/src/runtime.jl
```diff
@@ -181,16 +175,21 @@ function task_executor(ray_function, returns_ptr, task_args_ptr, task_name,
...
         end
+        args, kwargs = recover_args(flattened)
 
-        arg_string = join(string.("::", typeof.(args)), ", ")
-        @info "Calling $func($arg_string)"
+        @info begin
+            args_str = join([string("::", typeof(arg)) for arg in args], ", ")
+            kwargs_str = join([string(k, "::", typeof(v)) for (k, v) in kwargs], ", ")
+            args_kwargs_str = join([args_str, kwargs_str], "; ")
+            "Calling $func($args_kwargs_str)"
```
kleinschmidt, line 192: this seems more natural to me:
```suggestion
            "Calling $func($args_str; $kwargs_str)"
```

Ray.jl/src/runtime.jl
```diff
@@ -181,16 +175,21 @@ function task_executor(ray_function, returns_ptr, task_args_ptr, task_name,
...
             io = IOBuffer(v)
             return deserialize(io)
         end
+        args, kwargs = recover_args(flattened)
 
-        arg_string = join(string.("::", typeof.(args)), ", ")
-        @info "Calling $func($arg_string)"
+        @info begin
+            args_str = join([string("::", typeof(arg)) for arg in args], ", ")
+            kwargs_str = join([string(k, "::", typeof(v)) for (k, v) in kwargs], ", ")
```
kleinschmidt, line 190: I liked the broadcasting but maybe it's too clever?  or doesn't work with the kv pairs I guess...  if you're gonna use generators might as well make them fully lazy to not allocate teh intermediate arrays

```suggestion
            args_str = join((string("::", typeof(arg)) for arg in args), ", ")
            kwargs_str = join((string(k, "::", typeof(v)) for (k, v) in kwargs), ", ")
```

kleinschmidt (approved):
only extremely nitpicky comments! otherwise looks good

Ray.jl/src/runtime.jl
omus: I agree with you. I meant to get back to this as what I was trying to do was avoid having the "; " present when no keywords were used. I'll update the logic here to be:
```julia
isempty(kwargs_str) || (args_str *= "; " * kwargs_str)
"Calling $func($args_str)"
```

kleinschmidt: could make `kwargs_str = isempty(kwargs) ? "" : string("; ", ...)`

omus: Unfortunately with that we end up with `$func($args_str$kwargs_str)`

omus: Ended up with this:
```julia
param_str = join((string("::", typeof(arg)) for arg in args), ", ")
if !isempty(kwargs)
    param_str *= "; "
    param_str *= join((string(k, "::", typeof(v)) for (k, v) in kwargs), ", ")
end
"Calling $func($param_str)"
```
Having `param_str *= "; "` on a separate line just avoids hitting the line length limit

PR #51, opened 2023-08-28
Silent CxxWrap finalizer

omus:
Disabling the output on our CxxWrap finalizer which let's us know which objects are being deleted. It's still useful code when encountering segmentation faults which is why I've opted to comment the code out for now instead of delete it.


kleinschmidt (approved):
should we comment out the whole struct?

PR #65, opened 2023-08-29
Grab bag of minor changes

omus:
- Create an internal `_put` function which creates a `LocalMemoryBuffer`
- Move the `_submit_task` C++ function closer to the `task_executor_callback` as they are closely related and typically are changed at the same time
- Rename `rayjll` to `ray_jll`

### DIFF
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -6,14 +6,14 @@ the `data` with [`Ray.get`](@ref).
 """
 function put(data)
     buffer = _put(data)
-    return rayjll.put(buffer)
+    return ray_jll.put(buffer)
 end
 
 function _put(data)
     bytes = Vector{UInt8}()
     io = IOBuffer(bytes; write=true)
     serialize(io, data)
-    return rayjll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
+    return ray_jll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
 end
 
 """
@@ -26,8 +26,8 @@ if run in an `@async` task.
 If the task that generated the `ObjectID` failed with a Julia exception, the
 captured exception will be thrown on `get`.
 """
-get(oid::rayjll.ObjectIDAllocated) = _get(take!(rayjll.get(oid)))
-get(obj::SharedPtr{rayjll.RayObject}) = _get(take!(rayjll.GetData(obj[])))
+get(oid::ray_jll.ObjectIDAllocated) = _get(take!(ray_jll.get(oid)))
+get(obj::SharedPtr{ray_jll.RayObject}) = _get(take!(ray_jll.GetData(obj[])))
 get(x) = x
 
 function _get(data::Vector{UInt8})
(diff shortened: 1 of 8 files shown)


Ray.jl/src/object_store.jl
```diff
@@ -5,13 +5,15 @@ Store `data` in the object store. Returns an object reference which can used to
...
+function _put(data)
     bytes = Vector{UInt8}()
     io = IOBuffer(bytes; write=true)
     serialize(io, data)
-    buffer_ptr = Ptr{Nothing}(pointer(bytes))
-    buffer_size = sizeof(bytes)
-    buffer = rayjll.LocalMemoryBuffer(buffer_ptr, buffer_size, true)
-    return rayjll.put(buffer)
+    return ray_jll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
 end
```
kleinschmidt, line 17: I sorta feel like this should be a constructor for `LocalMemoryBuffer` (or `to_local_memory_buffer`) in the jll

Ray.jl/src/runtime.jl
```diff
@@ -225,17 +225,14 @@ function task_executor(ray_function, returns_ptr, task_args_ptr, task_name,
 
     # TODO: support multiple return values
     # https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/issues/54
-    buffer_data = Vector{UInt8}(sprint(serialize, result))
-    buffer_size = sizeof(buffer_data)
-    buffer = rayjll.LocalMemoryBuffer(buffer_data, buffer_size, true)
-    push!(returns, buffer)
+    push!(returns, _put(result))
```
kleinschmidt, line 228: yeah, this to me feels a bit misleading to call it "put", since all it's doing is serializing to a buffer, not actually doing a put...

Ray.jl/src/object_store.jl
omus: The problem is the `serialize` component. The `ray_jll` module doesn't care about serialization at all and you can put any data within a `ray_jll.LocalMemoryBuffer`. Maybe I'll just make this function `to_serialized_buffer`.

kleinschmidt: yeah. on the other hand, there's really only one thing you can put in a local memory buffer and that's a vector of bytes, and serialization seems like a good fallback for how to turn something into bytes...

kleinschmidt: I'm not too bothered about where this goes though as long as the name is clearer

Ray.jl/src/object_store.jl
```diff
@@ -26,13 +16,24 @@ if run in an `@async` task.
...
     # https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/issues/58
     result isa RayRemoteException ? throw(result) : return result
 end
+
+function to_serialized_buffer(data)
+    bytes = Vector{UInt8}()
+    io = IOBuffer(bytes; write=true)
+    serialize(io, data)
+    return ray_jll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
+end
```
kleinschmidt, line 35: another thing I'm not sure about is whether the LocalMemoryBuffer created this way will prevent julia GC from removing bytes; I _think_ so but I'm not 100% sure...

kleinschmidt (approved):

PR #66, opened 2023-08-29
support resource requests and use CPU: 1 for default tasks

kleinschmidt:
Need to do some slightly ugly work-arounds for lack of `std::unordered_map` support in CxxWrap, but otherwise pretty straightforward.

TODO: 
- [x] tests


kleinschmidt:
(this despite the fact that memory is one of the things that the python remote function special cases: https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/utils.py#L367-L407)

Ray.jl/test/task.jl
```diff
@@ -176,3 +176,17 @@ end
...
+
+@testset "task resource requests" begin
+    function gimme_resources()
+        resources = Ray.ray_jll.get_task_required_resources()
+        ks = Ray.ray_jll._keys(resources)
+        return Dict(String(k) => float(Ray.ray_jll._getindex(resources, k)) for k in ks)
+    end
+
+    default_resources = Ray.get(submit_task(gimme_resources, ()))
+    @test default_resources["CPU"] == 1.0
```
glennmoy, line 188: what are the other keys that can be set?

Ray.jl/src/runtime.jl
```diff
@@ -155,7 +155,8 @@ end
 initialize_coreworker_driver(args...) = ray_jll.initialize_coreworker_driver(args...)
 
 function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
-                     runtime_env::Union{RuntimeEnv,Nothing}=nothing)
+                     runtime_env::Union{RuntimeEnv,Nothing}=nothing,
+                     resources::Dict{String,Float64}=Dict("CPU" => 1.0))
```
glennmoy, line 159: if we don't set a fallback here does the job error? what happens to the other `options.resources` that aren't set?

Ray.jl/test/task.jl
kleinschmidt: great question!  I tried using `memory` since that's special cased in the python remote function wrapper but it lead to tasks not being schedulable at all.  in theory you can use any key that you want, but your nodes have to declare that they have that resource for tasks to be scheduled there

Ray.jl/src/runtime.jl
kleinschmidt: if you leave it blank, there are no resource requests, which means that ray is then free to over-schedule the CPU.  python uses `CPU: 1` as teh default.  I think there _are_ some concurrency limits on the raylet side (I've seen it spin up more workers than CPUs locally , but not _all_ the tasks get scheduled at once)

Ray.jl/test/task.jl
kleinschmidt: https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/remote_function.py#L153-L164

Ray.jl/src/runtime.jl
omus: Do you know where that default is set? From what I've seen it's left undefined

kleinschmidt: not sure, but the logs definitely show them with CPU: 1

Ray.jl/test/task.jl
kleinschmidt: https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/raylet/scheduling/scheduling_ids.h#L42-L45

omus (approved):
Looks great

Ray.jl/test/task.jl
kleinschmidt: WAIT: https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/protobuf/runtime_env_agent.proto#L28-L30

seems to imply that the units of memory request are GiB not bytes 🤦

kleinschmidt: ...but then there are things liek this in the tests:

https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/tests/test_basic.py#L610

which suggests that the requests on teh user side are in bytes

kleinschmidt: near as I can tell, the ptyhon remote function is doing exactly what we're doing: converting things to float adn dumping them into a dict in the task options: https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_raylet.pyx#L2899-L2909

PR #71, opened 2023-08-30
set logs_dir in options and default to file logging

kleinschmidt:
This fixes the issue where all teh core worker logs (including GCS event stats every 1s) were being written to stderr.

I did a tiny bit of refactoring to allow the session dir and logs dir to be passed in as arguments to `Ray.init()`.  Logging to stderr can be turned back on by either passing in `logs_dir=""`, or by setting `RAY_LOG_TO_STDERR=1` (which matches the python driver).  not totally wedded to that but figured it was good to be consistent.

---
  - https://app.asana.com/0/0/1205360251811528

### DIFF
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -60,7 +60,7 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
 
     # resolve symlink
     session_dir = realpath(session_dir)
-    log_to_stderr = get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
+    log_to_stderr = Base.get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
     logs_dir = log_to_stderr ? "" : joinpath(session_dir, "logs")
 
     if isnothing(runtime_env)
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -46,9 +46,14 @@ const GLOBAL_STATE_ACCESSOR = Ref{ray_jll.GlobalStateAccessor}()
 # https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/ray_constants.py#L198
 const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
 
+function default_log_dir(session_dir)
+    redirect_logs = Base.get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
+    return redirect_logs ? "" : realpath(joinpath(session_dir, "logs"))
+end
+
 function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
               session_dir="/tmp/ray/session_latest",
-              logs_dir=joinpath(realpath(session_dir), "logs"))
+              logs_dir=default_log_dir(session_dir))
     # XXX: this is at best EXREMELY IMPERFECT check.  we should do something
     # more like what hte python Worker class does, getting node ID at
     # initialization and using that as a proxy for whether it's connected or not
@@ -91,8 +96,6 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
     job_config = JobConfig(RuntimeEnvInfo(runtime_env))
     serialized_job_config = _serialize(job_config)
 
-    log_to_stderr = Base.get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
-    logs_dir = log_to_stderr ? "" : logs_dir
     ray_jll.initialize_driver(args..., job_id, logs_dir, serialized_job_config)
     atexit(ray_jll.shutdown_driver)
 
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -41,7 +41,13 @@ This is set during `init` and used there to get the Job ID for the driver.
 """
 const GLOBAL_STATE_ACCESSOR = Ref{ray_jll.GlobalStateAccessor}()
 
-function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing)
+# env var to control whether logs are sent do stderr or to file.  if "1", sent
+# to stderr; otherwise, will be sent to files in `/tmp/ray/session_latest/logs/`
+# https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/ray_constants.py#L198
+const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
+
+function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
+              session_dir="/tmp/ray/session_latest")
     # XXX: this is at best EXREMELY IMPERFECT check.  we should do something
     # more like what hte python Worker class does, getting node ID at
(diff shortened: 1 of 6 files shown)


Ray.jl/test/runtime.jl
```diff
@@ -0,0 +1,39 @@
...
+
+    @testset "custom logs_dir" begin
+        mktempdir() do logs_dir
+            cp("/tmp/ray/session_latest/logs/raylet.out", joinpath(logs_dir, "raylet.out"))
+            code = quote
+                using Ray
+                Ray.init(; logs_dir=$logs_dir)
+            end
+            cmd = `$(Base.julia_cmd()) --project=$(Ray.project_dir()) -e $code`
+            run(cmd)
```
glennmoy, line 16: it might be good to test that - given a `logs_dir` - we _dont'_ get ray logs printed to stderr?

Ray.jl/test/runtime.jl
```diff
@@ -0,0 +1,39 @@
+@testest "control logs_dir" begin
+    @testest "defaul logs_dir" begin
+        # we're running this after `init` has been called, so logs should exist already:
+        logs = readdir("/tmp/ray/session_latest/logs"; join=true)
+        @test any(contains("julia-core-driver"), logs)
+    end
+
+    @testset "custom logs_dir" begin
+        mktempdir() do logs_dir
+            cp("/tmp/ray/session_latest/logs/raylet.out", joinpath(logs_dir, "raylet.out"))
```
glennmoy, line 10: why do we need to copy the `raylet.out` over?

Ray.jl/src/runtime.jl
```diff
@@ -81,7 +91,9 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing)
     job_config = JobConfig(RuntimeEnvInfo(runtime_env))
     serialized_job_config = _serialize(job_config)
 
-    ray_jll.initialize_driver(args..., job_id, serialized_job_config)
+    log_to_stderr = Base.get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
+    logs_dir = log_to_stderr ? "" : logs_dir
```
glennmoy, line 95: so if I'm reading this right...it's not enough to provide a `logs_dir` to get the logs directed to a file...as this can be totally ignored by an environment variable? which seems...convoluted/surprising?

(note: I know this is a Ray thing I'm just confused why the ENV var can obviate the (ostensible) purpose of a function argument...unless the ENV var controls much more than this and there's more going on than I'm aware...)

Ray.jl/test/runtime.jl
glennmoy: update: _non-empty_ `logs_dir` and ENV var set to "0"

Ray.jl/src/runtime.jl
kleinschmidt: yeah, this might not be a great idea, I was just trying to be consistent with the python implementation (which checks this env var)

Ray.jl/test/runtime.jl
```diff
@@ -0,0 +1,58 @@
...
+            code = quote
+                using Ray
+                Ray.init(; logs_dir=$logs_dir)
+            end
+            cmd = `$(Base.julia_cmd()) --project=$(Ray.project_dir()) -e $code`
+            err = IOBuffer()
+            run(pipeline(cmd; stderr=err))
```
omus, line 17: We could probably iterate on `process_eval` to support this

Ray.jl/test/runtime.jl
```diff
@@ -0,0 +1,58 @@
+@testset "control logs_dir" begin
+    @testset "defaul logs_dir" begin
```
omus, line 2:
```suggestion
    @testset "default logs_dir" begin
```

Ray.jl/test/runtime.jl
omus: Why not use a symlink for this purpose? Seems clearer

Ray.jl/src/runtime.jl
omus: I think it's better if this option is controllable via arguments from the `Ray.init` call and the `ENV` setting just modifies the default. Something along the lines of:

```julia
function default_log_dir(session_dir)
    redirect_logs = Base.get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
    return redirect_logs ? "" : joinpath(session_dir, "logs")
end

function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
              session_dir="/tmp/ray/session_latest",
              logs_dir=default_log_dir(session_dir))
    ...
end
```

omus: The difference with my proposal is that `logs_dir` takes precedence over `ENV["RAY_LOG_TO_STDERR"]` which currently isn't true

Ray.jl/test/runtime.jl
kleinschmidt: eh, it's throwaway in a tempdir for a test.  these files _can_ get big when they're running for a long time but i don't think that's likely to have a meaningful impact in this test suite

kleinschmidt: just gonna leave it as cp for now to keep this moving.

Ray.jl/src/runtime.jl
```diff
@@ -41,7 +41,19 @@ This is set during `init` and used there to get the Job ID for the driver.
...
 
-function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing)
+# env var to control whether logs are sent do stderr or to file.  if "1", sent
+# to stderr; otherwise, will be sent to files in `/tmp/ray/session_latest/logs/`
+# https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/ray_constants.py#L198
+const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
+
+function default_log_dir(session_dir)
+    redirect_logs = Base.get(ENV, LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE, "0") == "1"
+    return redirect_logs ? "" : realpath(joinpath(session_dir, "logs"))
```
omus, line 51: Is `realpath` actually required?

Ray.jl/test/runtime.jl
```diff
@@ -0,0 +1,82 @@
...
+        code = quote
+            using Ray
+            Ray.init(; logs_dir="")
+        end
+        cmd = `$(Base.julia_cmd()) --project=$(Ray.project_dir()) -e $code`
+        out = IOBuffer()
+        err = IOBuffer()
+        run(pipeline(cmd; stdout=out, stderr=err))
+        logs = String(take!(err))
+        @test contains(logs, "Constructing CoreWorkerProcess")
```
omus, line 41:
```suggestion
        stderr_logs = String(take!(err))
        @test contains(stderr_logs, "Constructing CoreWorkerProcess")
```

Ray.jl/test/runtime.jl
```diff
@@ -0,0 +1,82 @@
...
+            using Ray
+            ENV[Ray.LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE] = "1"
+            Ray.init()
+        end
+        cmd = `$(Base.julia_cmd()) --project=$(Ray.project_dir()) -e $code`
+        out = IOBuffer()
+        err = IOBuffer()
+        run(pipeline(cmd; stdout=out, stderr=err))
+        logs = String(take!(err))
+        @test contains(logs, "Constructing CoreWorkerProcess")
```
omus, line 55:
```suggestion
        stderr_logs = String(take!(err))
        @test contains(stderr_logs, "Constructing CoreWorkerProcess")
```

omus (approved):
LGTM

Ray.jl/src/runtime.jl
omus: Worth adding a comment about this rational

PR #72, opened 2023-08-31
Support inlining task arguments

omus:
Should improve performance for Ray task calls which use small sized arguments.

### DIFF
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -178,7 +178,7 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
                      resources::Dict{String,Float64}=Dict("CPU" => 1.0))
     export_function!(FUNCTION_MANAGER[], f, get_current_job_id())
     fd = ray_jll.function_descriptor(f)
-    arg_oids = map(Ray.put, flatten_args(args, kwargs))
+    task_args = prepare_task_args(flatten_args(args, kwargs))
 
     serialized_runtime_env_info = if !isnothing(runtime_env)
         _serialize(RuntimeEnvInfo(runtime_env))
@@ -187,7 +187,7 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
     end
 
     return GC.@preserve args ray_jll._submit_task(fd,
-                                                  arg_oids,
+                                                  task_args,
                                                   serialized_runtime_env_info,
                                                   resources)
 end
@@ -199,19 +199,24 @@ function byte_serialize(data)
     return bytes
 end
 
+#=
+using Ray; import ray_core_worker_julia_jll as ray_jll
+Ray.prepare_task_args([1,2,3])
+=#
+
 # TODO: be smarter about handling flattened args
 # Adapted from `prepare_args_internal`:
 # https://github.com/ray-project/ray/blob/ray-2.5.1/python/ray/_raylet.pyx#L673
 function prepare_task_args(args)
     ray_config = ray_jll.RayConfigInstance()
     put_threshold = ray_jll.max_direct_call_object_size(ray_config)
+    total_inlined = 0
     rpc_inline_threshold = ray_jll.task_rpc_inlined_bytes_limit(ray_config)
-    core_worker = ray_jll.GetCoreWorker()
-    rpc_address = ray_jll.GetRpcAddress(core_worker)
+    # core_worker = ray_jll.GetCoreWorker()
+    rpc_address = ray_jll.GetRpcAddress()
 
     # TODO: put_arg_call_site
 
-    #=
     task_args = []
     for arg in args
         # if arg isa ObjectRef
@@ -231,23 +236,15 @@ function prepare_task_args(args)
 
         if serialized_arg_size <= put_threshold && serialized_arg_size + total_inlined <= rpc_inline_threshold
             buffer = ray_jll.LocalMemoryBuffer(serialized_arg, serialized_arg_size, true)
-            push!(task_args, TaskArgByValue(serialized_arg))
+            push!(task_args, ray_jll.TaskArgByValue(ray_jll.RayObject(buffer)))
             total_inlined += serialized_arg_size
         else
-            push!(task_args, TaskArgByReference(put_id, rpc_address, nullptr))
-
-
-        to_serialized_buffer
-
-        if
-
-
-
-
+            error("not implemented")
+            # push!(task_args, TaskArgByReference(put_id, rpc_address, nullptr))
+        end
     end
-    =#
 
-
+    return task_args
 end
 
 function task_executor(ray_function, returns_ptr, task_args_ptr, task_name,
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -192,6 +192,64 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
                                                   resources)
 end
 
+function byte_serialize(data)
+    bytes = Vector{UInt8}()
+    io = IOBuffer(bytes; write=true)
+    serialize(io, data)
+    return bytes
+end
+
+# TODO: be smarter about handling flattened args
+# Adapted from `prepare_args_internal`:
+# https://github.com/ray-project/ray/blob/ray-2.5.1/python/ray/_raylet.pyx#L673
+function prepare_task_args(args)
+    ray_config = ray_jll.RayConfigInstance()
+    put_threshold = ray_jll.max_direct_call_object_size(ray_config)
+    rpc_inline_threshold = ray_jll.task_rpc_inlined_bytes_limit(ray_config)
+    core_worker = ray_jll.GetCoreWorker()
+    rpc_address = ray_jll.GetRpcAddress(core_worker)
+
+    # TODO: put_arg_call_site
+
+    #=
+    task_args = []
+    for arg in args
+        # if arg isa ObjectRef
+        #     oid = arg.oid
+        #     op_status = ray_jll.GetOwnerAddress(core_worker, oid, owner_address)
+        #             c_arg, &c_owner_address)
+        #     check_status(op_status)
+        #     push!(task_args,
+        #         unique_ptr[CTaskArg](new CTaskArgByReference(
+        #             c_arg,
+        #             c_owner_address,
+        #             arg.call_site())))
+        # end
+
+        serialized_arg = byte_serialize(arg)
+        serialized_arg_size = sizeof(serialized_arg)
+
+        if serialized_arg_size <= put_threshold && serialized_arg_size + total_inlined <= rpc_inline_threshold
+            buffer = ray_jll.LocalMemoryBuffer(serialized_arg, serialized_arg_size, true)
+            push!(task_args, TaskArgByValue(serialized_arg))
+            total_inlined += serialized_arg_size
+        else
+            push!(task_args, TaskArgByReference(put_id, rpc_address, nullptr))
+
+
+        to_serialized_buffer
+
+        if
+
+
+
+
+    end
+    =#
+
+
+end
+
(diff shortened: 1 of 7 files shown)


omus:
I've been attempting to construct the `TaskArgByValue` and `TaskArgByReference` instances in Julia and then pass them directly into
`submit_task`. However, this is challenging as I need to pass in a `std::vector<std::unique_ptr<TaskArg>>`. 

Compiling this example fails:
```cpp
size_t demo(const std::vector<std::unique_ptr<std::string>> &vector) {
    return vector.size();
}

JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
{
    mod.method("demo", &demo);
}
```

Failure:
```
bazel-out/darwin_arm64-opt/bin/external/libcxxwrap_julia/_virtual_includes/headers/jlcxx/stl.hpp:173:89: error: object of type 'std::__deque_base<std::unique_ptr<std::string>, std::allocator<std::unique_ptr<std::string>>>::value_type' (aka 'std::unique_ptr<std::string>') cannot be assigned because its copy assignment operator is implicitly deleted
    wrapped.method("cxxsetindex!", [](WrappedT& v, const T& val, cxxint_t i) { v[i - 1] = val; });
```
As reported in: https://github.com/JuliaInterop/CxxWrap.jl/issues/370

I've worked around this issue for now by creating a `StdVector` of `TaskArg` pointers and then converting these pointers to `std::unique_ptr` inside of the C++ code.

One addition challenge was around `TaskArg` being an abstract C++ class. In particular attempting to use a Julia `Vector{CxxRef{TaskArg}}` and pushing `CxxRef{TaskArgByValueAllocated}` values fails as it tries to convert to an abstract type. Using a `Vector{Any}` does work but when converting to a `StdVector` we get a `StdVector{Any}` which is an issue for calling the C++ code. The solution here was to start with an empty `StdVector` allocated on the Julia side and then push elements to it. This works but required addtional work since CxxWrap was automatically dereferencing my values.

omus:
I merged in #74 here as there was some overlap which makes the diff messy (TIL). Once we merge #74 I'll clean this PR up. Besides that this PR is ready for review.

omus:
I've rebased these changes to no longer depend on #74

Ray.jl/src/runtime.jl
```diff
@@ -173,23 +173,76 @@ end
...
+    return task_args
+end
+
+function transform_task_args(task_args)
+    task_arg_ptrs = StdVector{CxxPtr{ray_jll.TaskArg}}()
+    for task_arg in task_args
+        push!(task_arg_ptrs, CxxRef(task_arg))
+    end
+    return task_arg_ptrs
 end
```
omus, line 246: One liner doesn't currently work:
```
julia> CxxRef.(convert.(ray_jll.TaskArg, task_args))
3-element Vector{CxxRef{ray_core_worker_julia_jll.TaskArg}}:
 CxxRef{ray_core_worker_julia_jll.TaskArg}(Ptr{ray_core_worker_julia_jll.TaskArg} @0x0000600000051300)
 CxxRef{ray_core_worker_julia_jll.TaskArg}(Ptr{ray_core_worker_julia_jll.TaskArg} @0x000060000004de60)
 CxxRef{ray_core_worker_julia_jll.TaskArg}(Ptr{ray_core_worker_julia_jll.TaskArg} @0x000060000004d0e0)

julia> StdVector{CxxPtr{ray_jll.TaskArg}}(CxxRef.(convert.(ray_jll.TaskArg, task_args)))
ERROR: MethodError: no method matching StdVector{CxxPtr{ray_core_worker_julia_jll.TaskArg}}(::Vector{CxxRef{ray_core_worker_julia_jll.TaskArg}})

Closest candidates are:
  StdVector{CxxPtr{ray_core_worker_julia_jll.TaskArg}}()

Stacktrace:
 [1] top-level scope
   @ REPL[15]:1
```

deps/wrapper.cc
```diff
@@ -541,4 +587,39 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
...
+        .method("RayConfigInstance", &RayConfig::instance)
+        .method("max_direct_call_object_size", [](RayConfig &config) {
+            return config.max_direct_call_object_size();
+        })
+        .method("task_rpc_inlined_bytes_limit", [](RayConfig &config) {
+            return config.task_rpc_inlined_bytes_limit();
+        })
+        .method("record_ref_creation_sites", [](RayConfig &config) {
+            return config.record_ref_creation_sites();
+        });
```
kleinschmidt, line 601: why do these need to be lambdas? seem pretty straightforward...

src/wrappers/any.jl
```diff
@@ -148,12 +148,17 @@ function Base.push!(v::CxxPtr{StdVector{T}}, el::T) where T <: SharedPtr{LocalMe
...
+    return v
+end
+
 # XXX: Need to convert julia vectors to StdVector and build the
 # `std::unordered_map` for resources. This function helps us avoid having
 # CxxWrap as a direct dependency in Ray.jl
-function _submit_task(fd, oids::AbstractVector, serialized_runtime_env_info, resources)
-    # https://github.com/JuliaInterop/CxxWrap.jl/issues/367
-    args = isempty(oids) ? StdVector{ObjectID}() : StdVector(oids)
+function _submit_task(fd, args, serialized_runtime_env_info, resources::AbstractDict)
```
kleinschmidt, line 161: any reason to remove the `AbstractVector` restriction here?

deps/wrapper.cc
```diff
@@ -149,9 +149,11 @@ ObjectID _submit_task(const ray::JuliaFunctionDescriptor &jl_func_descriptor,
     ray::FunctionDescriptor func_descriptor = std::make_shared<ray::JuliaFunctionDescriptor>(jl_func_descriptor);
     RayFunction func(Language::JULIA, func_descriptor);
 
+    // TODO: Passing in a `std::vector<std::unique_ptr<TaskArg>>` from Julia may currently be impossible due to:
+    // https://github.com/JuliaInterop/CxxWrap.jl/issues/370
     std::vector<std::unique_ptr<TaskArg>> args;
-    for (auto & obj_id : object_ids) {
-        args.emplace_back(new TaskArgByReference(obj_id, worker.GetRpcAddress(), /*call-site*/""));
+    for (auto &task_arg : task_args) {
+        args.emplace_back(task_arg);
```
kleinschmidt, line 156: since the element type is `unique_ptr` already I'm assuming this will auto-cast to `unique_ptr`?

kleinschmidt: ah wait I always forget, `emplace_back` constructs and pushes...

Ray.jl/test/runtime.jl
```diff
@@ -78,5 +78,41 @@
...
+    ray_config = ray_jll.RayConfigInstance()
+    put_threshold = ray_jll.max_direct_call_object_size(ray_config)
+    rpc_inline_threshold = ray_jll.task_rpc_inlined_bytes_limit(ray_config)
+    serialization_overhead = begin
+        sizeof(Ray.serialize_to_bytes(:_ => zeros(UInt8, put_threshold))) - put_threshold
+    end
+
+    @testset "put threshold" begin
+        a = :_ => zeros(UInt8, put_threshold - serialization_overhead)
+        b = :_ => zeros(UInt8, put_threshold - serialization_overhead + 1)
```
kleinschmidt, line 93: was really scratching my head here until I remembered this is how we represent positional arguments.  thought it was some weird `fill` syntax.  maybe add a comment to clarify?

```suggestion
        # positional arguments to tasks have key `:_`
        a = :_ => zeros(UInt8, put_threshold - serialization_overhead)
        b = :_ => zeros(UInt8, put_threshold - serialization_overhead + 1)
```

Ray.jl/src/runtime.jl
```diff
@@ -173,23 +173,76 @@ end
...
+    rpc_address = ray_jll.GetRpcAddress()
+
+    total_inlined = 0
+    task_args = Any[]
+    for arg in args
+        # Note: The Python `prepare_args_internal` function checks if the `arg` is an
+        # `ObjectRef` and in that case uses the object ID to directly make a
+        # `TaskArgByReference`. However, as the `args` here are flattened the `arg` will
+        # always be a `Pair` (or a list in Python). I suspect this Python code path just
+        # dead code so we'll exclude it from ours.
```
kleinschmidt, line 215: _should_ we be doing this though?  the values will be automatically dereferenced on the the task execution side so I'm not so sure; then again, if we _don't_ do it then users may be surprised if they _do_ have to de-reference in their work functions.

Ray.jl/src/runtime.jl
```diff
@@ -173,23 +173,76 @@ end
...
+
+        serialized_arg = serialize_to_bytes(arg)
+        serialized_arg_size = sizeof(serialized_arg)
+        buffer = ray_jll.LocalMemoryBuffer(serialized_arg, serialized_arg_size, true)
+
+        # Inline arguments which are small and if there is room
+        task_arg = if (serialized_arg_size <= put_threshold &&
+            serialized_arg_size + total_inlined <= rpc_inline_threshold)
```
kleinschmidt, line 223:
```suggestion
        task_arg = if (serialized_arg_size <= put_threshold &&
                       serialized_arg_size + total_inlined <= rpc_inline_threshold)
```

Ray.jl/src/runtime.jl
omus: https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/issues/80

kleinschmidt (approved):
I don't have any objections here; a few tiny suggestions and clarification questions but otherwise LGTM

Ray.jl/src/runtime.jl
omus: The Python code I'm referring to is attempting to avoid creating a new `ObjectRef` here if the user has passed in one as an argument. In all cases Python will dereference the arguments inside the called task. What I was specifically calling out here is that due to `flatten_args` this optimization will never occur as an `arg` is never a `ObjectRef`. At best it would be a `list` containing an `ObjectRef` as the second value.

Ray.jl/test/runtime.jl
kleinschmidt: or use `flatten_args`

kleinschmidt: actually, I think I like using `flatten_args` better here

omus: I'll add this further up as the first occurrence is by `serialization_overhead`

deps/wrapper.cc
omus: > since the element type is `unique_ptr` already I'm assuming this will auto-cast to `unique_ptr`?

Just to avoid any possible confusion the `task_arg` element is just a raw pointer and is not already a `unique_ptr`

src/wrappers/any.jl
omus: The `args` passed in here are a `StdVector` so if we'd leave the type assertion in we'd have to update it. Mainly this assertion was dropped as this method should only be called if the `resources` argument is a `AbstractDict`

Ray.jl/src/runtime.jl
```diff
@@ -173,23 +173,76 @@ end
...
+        push!(task_args, task_arg)
+    end
+
+    return task_args
+end
+
+function transform_task_args(task_args)
+    task_arg_ptrs = StdVector{CxxPtr{ray_jll.TaskArg}}()
+    for task_arg in task_args
+        push!(task_arg_ptrs, CxxRef(task_arg))
```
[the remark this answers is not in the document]
omus: So this is calling our custom `push!` which calls `ray_jll._push_back`. I believe this is specifically due to me defining:
```cpp
void _push_back(std::vector<TaskArg *> &vector, TaskArg &el) {
    vector.push_back(&el);
}
```
instead of:
```cpp
void _push_back(std::vector<TaskArg *> &vector, TaskArg *el) {
    vector.push_back(el);
}
```
I'll update the code so we can just pass in a `CxxPtr` here

PR #74, opened 2023-08-31
Pass `CoreWorker` into wrapped C++ code

### LINKED ISSUE #61: Avoid over-calling GetCoreWorker()
probably makes more sense to have a global worker rather than calling `GetCoreWorker()` over and over again in, e.g., `GetCurrentJobId`, `put` and `get`.

omus:
Fix #61

### DIFF
--- a/Ray.jl/src/runtime.jl
+++ b/Ray.jl/src/runtime.jl
@@ -108,15 +108,6 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
     return nothing
 end
 
-const CORE_WORKER = Ref{ray_jll.CoreWorker}()
-
-function GetCoreWorker()
-    if !isassigned(CORE_WORKER)
-        CORE_WORKER[] = ray_jll.GetCoreWorker()[]
-    end
-    return CORE_WORKER[]
-end
-
 # TODO: Python Ray returns a string:
 # https://docs.ray.io/en/latest/ray-core/api/doc/ray.runtime_context.RuntimeContext.get_job_id.html
 
@@ -126,14 +117,14 @@ end
 Get the current job ID for this worker or driver. Job ID is the id of your Ray drivers that
 create tasks.
 """
-get_job_id() = ray_jll.ToInt(ray_jll.GetCurrentJobId(GetCoreWorker()))
+get_job_id() = ray_jll.ToInt(ray_jll.GetCurrentJobId(ray_jll.GetCoreWorker()))
 
 """
     get_task_id() -> String
 
 Get the current task ID for this worker in hex format.
 """
-get_task_id() = String(ray_jll.Hex(ray_jll.GetCurrentTaskId(GetCoreWorker())))
+get_task_id() = String(ray_jll.Hex(ray_jll.GetCurrentTaskId(ray_jll.GetCoreWorker())))
 
 function parse_ray_args_from_raylet_out(session_dir)
     #=
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -176,7 +176,7 @@ ObjectID _submit_task(const ray::JuliaFunctionDescriptor &jl_func_descriptor,
     return ObjectRefsToIds(return_refs)[0];
 }
 
-ray::core::CoreWorker &GetCoreWorker() {
+ray::core::CoreWorker &_GetCoreWorker() {
     return CoreWorkerProcess::GetCoreWorker();
 }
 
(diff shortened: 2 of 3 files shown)


Ray.jl/src/runtime.jl
```diff
@@ -108,16 +108,33 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
...
-get_current_job_id() = ray_jll.ToInt(ray_jll.GetCurrentJobId())
+const CORE_WORKER = Ref{CxxRef{ray_jll.CoreWorker}}()
+
+function GetCoreWorker()
+    return ray_jll.GetCoreWorker()
+    # if !isassigned(CORE_WORKER)
+    #     CORE_WORKER[] = ray_jll.GetCoreWorker()
+    # end
+    # return CORE_WORKER[]
+end
```
omus, line 119: Using the write-once version of this function results in most of the Ray.jl test suite failing with:
```
C++ object of type N3ray4core10CoreWorkerE was deleted
```
I'll try a couple other things

omus: Interesting. If I change our global reference to be a `Ref{ray_jll.CoreWorker}` then things work fine. It seems that the `CxxRef` doesn't hold onto the allocated memory.

omus:
CI failure will be fixed by #75

deps/wrapper.cc
```diff
@@ -533,12 +511,20 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
...
         .method("Disconnect", &ray::gcs::GlobalStateAccessor::Disconnect);
 
-    mod.method("report_error", &report_error);
+    // https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/core_worker.h#L284
+    mod.add_type<ray::core::CoreWorker>("CoreWorker")
+        .method("GetCurrentJobId", &ray::core::CoreWorker::GetCurrentJobId)
+        .method("GetCurrentTaskId", &ray::core::CoreWorker::GetCurrentTaskId)
+        .method("put", &put)
+        .method("get", &get)
+        .method("submit_task", &submit_task)
```
kleinschmidt, line 520: it's not clear to me why we are exposing these are methods of this class when they're not defined that way.

kleinschmidt (commented):
I think my concern here is that it's not clear what the problem is that this is solving; it introduces some fragility by making the julia side keep track of teh C++ coreworker struct.  That _might_ be worth it if there's a clear benefit but from what I can tell right now, there isn't.

Ray.jl/src/runtime.jl
```diff
@@ -108,16 +108,32 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
...
+# TODO: Python Ray returns a string:
+# https://docs.ray.io/en/latest/ray-core/api/doc/ray.runtime_context.RuntimeContext.get_job_id.html
+
+"""
+    get_job_id() -> UInt32
+
+Get the current job ID for this worker or driver. Job ID is the id of your Ray drivers that
+create tasks.
+"""
+get_job_id() = ray_jll.ToInt(ray_jll.GetCurrentJobId(GetCoreWorker()))
```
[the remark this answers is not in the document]
omus: The Python Ray documentation uses this for the name: https://docs.ray.io/en/latest/ray-core/api/doc/ray.runtime_context.RuntimeContext.get_job_id.html

Ray.jl/src/runtime.jl
```diff
@@ -108,16 +108,32 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
...
-# keep it local
-get_current_job_id() = ray_jll.ToInt(ray_jll.GetCurrentJobId())
+const CORE_WORKER = Ref{ray_jll.CoreWorker}()
+
+function GetCoreWorker()
+    if !isassigned(CORE_WORKER)
+        CORE_WORKER[] = ray_jll.GetCoreWorker()[]
+    end
+    return CxxRef(CORE_WORKER[])
+end
```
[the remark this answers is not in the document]
omus: Originally the reason for this was to handle the `start_worker` case. In the scenario where `start_worker` is called we call the blocking call `initialize_worker` which is the point at which `CoreWorker` is instantiated. Doing this this way allows this case to work. It also ended up being useful when I changed the code to return a `CxxRef`.

deps/wrapper.cc
omus: I added these as methods of this class only to avoid headaches with moving these definitions around. The way we've set these up in Julia these methods are associated with this type so doing this just indicates the relationship here. Definitely optional

omus:
> I think my concern here is that it's not clear what the problem is that this is solving; it introduces some fragility by making the julia side keep track of teh C++ coreworker struct. That _might_ be worth it if there's a clear benefit but from what I can tell right now, there isn't.

Continuing on this thread the C++ code typically just calls `GetCoreWorker()` over and over so I don't think there's a major performance hit with doing this on the C++ side. However, if we did this in Julia we'd be allocating and deallocating over and over again so we would see a performance hit.

After implementing this PR I think what we should do is only pass in a `CoreWorker` for it's own class methods and just keep calling `GetCoreWorker` internally in the C++ code. For this to be performant however we'd need to keep the cached `Ray.GetCoreWorker` function. What do you think?

omus:
Last thing to do here is to move the caching `GetCoreWorker()` into the JLL.

omus:
RTM again

kleinschmidt (approved):
still a bit nervous about handlign this on the julia side but at least in teh JLL it's confined to a safe space.  and as discussed in huddle, if we're not handling it correctly, we don't really understand what we're doing here (in a general sense) so probably better to surface that.

PR #75, opened 2023-09-01
Fix "Deadline Exceeded" exception

omus:
Fixes this exception I was seeing regularly on my MacBook:
```julia
function manager: Error During Test at /Users/cvogt/.julia/dev/ray_core_worker_julia_jll/Ray.jl/test/function_manager.jl:23
  Test threw exception
  Expression: wait_for_function(fm, mfd, jobid; timeout_s = 1) == :timed_out
  Unknown: Deadline Exceeded
  Stacktrace:
   [1] Exists(arg1::Union{JuliaGcsClient, CxxWrap.CxxWrapCore.CxxRef{<:JuliaGcsClient}}, arg2::Union{AbstractString, CxxWrap.CxxWrapCore.ConstCxxRef{<:CxxWrap.StdLib.StdString}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.StdString}}, arg3::Union{AbstractString, CxxWrap.CxxWrapCore.ConstCxxRef{<:CxxWrap.StdLib.StdString}, CxxWrap.CxxWrapCore.CxxRef{<:CxxWrap.StdLib.StdString}}, arg4::Integer)
   [2] #3
     @ ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/function_manager.jl:100 [inlined]
   [3] timedwait(testcb::Ray.var"#3#4"{Int64, FunctionManager, String}, timeout::Int64; pollint::Float64)
     @ Base ./asyncevent.jl:320
   [4] timedwait
     @ ./asyncevent.jl:311 [inlined]
   [5] #wait_for_function#2
     @ ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/function_manager.jl:97 [inlined]
   [6] macro expansion
     @ ~/Development/Julia/aarch64/1.9/usr/share/julia/stdlib/v1.9/Test/src/Test.jl:478 [inlined]
   [7] macro expansion
     @ ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/test/function_manager.jl:23 [inlined]
   [8] macro expansion
     @ ~/Development/Julia/aarch64/1.9/usr/share/julia/stdlib/v1.9/Test/src/Test.jl:1498 [inlined]
   [9] top-level scope
     @ ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/test/function_manager.jl:2
```

### DIFF
--- a/Ray.jl/src/function_manager.jl
+++ b/Ray.jl/src/function_manager.jl
@@ -90,7 +90,7 @@ function export_function!(fm::FunctionManager, f, job_id=get_current_job_id())
     end
 end
 
-function wait_for_function(fm::FunctionManager, fd::ray_jll.JuliaFunctionDescriptor,
+function timedwait_for_function(fm::FunctionManager, fd::ray_jll.JuliaFunctionDescriptor,
                            job_id=get_current_job_id(); timeout_s=10)
     key = function_key(fd, job_id)
     status = try
--- a/Ray.jl/src/function_manager.jl
+++ b/Ray.jl/src/function_manager.jl
@@ -91,13 +91,17 @@ function export_function!(fm::FunctionManager, f, job_id=get_current_job_id())
 end
 
 function wait_for_function(fm::FunctionManager, fd::ray_jll.JuliaFunctionDescriptor,
-                           job_id=get_current_job_id();
-                           pollint_s=0.01, timeout_s=10)
+                           job_id=get_current_job_id(); timeout=10)
     key = function_key(fd, job_id)
-    status = timedwait(timeout_s; pollint=pollint_s) do
-        # Set Exists to not block by using a negative timeout. If we set a non-negative
-        # timeout then this function can throw "Deadline Exceeded" exceptions.
-        ray_jll.Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, -1)
+    status = try
+        exists = ray_jll.Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, timeout)
+        exists ? :ok : :timed_out
+    catch e
+        if e isa ErrorException && contains(e.msg, "Deadline Exceeded")
+            return :timed_out
+        else
+            rethrow()
+        end
     end
     return status
 end
--- a/Ray.jl/src/function_manager.jl
+++ b/Ray.jl/src/function_manager.jl
@@ -95,9 +95,9 @@ function wait_for_function(fm::FunctionManager, fd::ray_jll.JuliaFunctionDescrip
                            pollint_s=0.01, timeout_s=10)
     key = function_key(fd, job_id)
     status = timedwait(timeout_s; pollint=pollint_s) do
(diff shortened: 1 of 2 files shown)


Ray.jl/src/function_manager.jl
```diff
@@ -95,9 +95,9 @@ function wait_for_function(fm::FunctionManager, fd::ray_jll.JuliaFunctionDescrip
                            pollint_s=0.01, timeout_s=10)
     key = function_key(fd, job_id)
     status = timedwait(timeout_s; pollint=pollint_s) do
-        # timeout the Exists query to the same timeout we use here so we don't
-        # deadlock.
-        ray_jll.Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, timeout_s)
+        # Set Exists to not block by using a negative timeout. If we set a non-negative
+        # timeout then this function can throw "Deadline Exceeded" exceptions.
+        ray_jll.Exists(fm.gcs_client, FUNCTION_MANAGER_NAMESPACE, key, -1)
```
omus, line 100: Alternatively we could update the `ray_jll.Exists` function to return `false` when we hit time timeout. Something like:
```cpp
if (!status.ok()) {
    std::string status_str = status.ToString();

    // Key doesn't exist before we hit the timeout
    if (timeout_ms >= 0 && status_str.find("Deadline Exceeded") != std::string::npos) {
        return false;
    }
    throw std::runtime_error(status_str);
}
```
This approach means we'd no longer need the `timedwait`

kleinschmidt (changes requested):
I think it's better to catch the error rather than possibly deadlocking.  we can't interrupt the Exists call IIUC so this is pretty dangerous (may never timeout).  I wouldn't want to merge this unless we're sure that the timeout is actually honored; not quite sure how to test that, maybe by connecting to a non-existent port?

omus:
@kleinschmidt were talking about what the behaviour of a `-1` timeout was. From my observations a `-1` returns immediately but Dave thought it would hang forever. I ended up looking into the Ray code and found this:
https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/gcs/gcs_client/gcs_client.cc#L164-L169

In learning more about gRPC deadlines the [TL;DR is: "Always set a deadline"](https://grpc.io/blog/deadlines/). With this in mind it makes sense to either set a deadline to be equal to the poll interval or just let the `InternalKVGet` code handle the timeout and remove `timedwait`. I've opted for the later.

Additionally, it seems strange to have `Exist` return `false` when a key doesn't exist or the function has timed out. Because of this I've decided to just to capture the exception on the Julia side. I don't love this setup but we should eventually move away from using the `PythonGcsClient` (https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/issues/76) which would grant us more control in handling things like the deadline.

kleinschmidt (approved):
I'm good with this except for the naming of `timeout`.

Also, we may want to name this to `timedwait_for_function` since that's really the interface we have here...up to you.

PR #78, opened 2023-09-01
add timeout to `ray_jll.get` and make `Ray.get` short-poll every 100ms

### LINKED ISSUE #69: `Ray.get` on a object ID that is not fulfilled locks julia
Can't be killed with Ctrl-C either.  Should somehow make the wait happen asynchronously if possible (can we put it in another thread?)

kleinschmidt:
This preserves the high-level behavior of `Ray.get` (blocks indefinitely until object is ready) while doing so in a way that does not hard lock the julia session.  Ultimately we'll also want to do something similar for `wait` (without actually trying to _fetch_ the results) but this is a fine stopgap for now.

---
  - https://app.asana.com/0/0/1205391426815967

### DIFF
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -34,14 +34,7 @@ function get(oid::ray_jll.ObjectIDAllocated)
     return get(ray_obj)
 end
 
-function get(ray_obj::SharedPtr{ray_jll.RayObject})
-    if isnull(ray_obj[])
-        return nothing
-    else
-        return _get(take!(ray_jll.GetData(ray_obj[])))
-    end
-end
-
+get(ray_obj::SharedPtr{ray_jll.RayObject}) = _get(take!(ray_jll.GetData(ray_obj[])))
 get(x; kwargs...) = x
 
 function _get(bytes)
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -14,7 +14,7 @@ end
 put(obj_ref::ObjectRef) = obj_ref
 
 """
-    Ray.get(object_id::ObjectIDAllocated; timeout_ms=-1)
+    Ray.get(object_id::ObjectIDAllocated)
 
 Retrieves the data associated with the `object_id` from the object store.  This
 method is blocking until the data is available in the local object store, even
@@ -23,10 +23,20 @@ if run in an `@async` task.
 If the task that generated the `ObjectID` failed with a Julia exception, the
 captured exception will be thrown on `get`.
 """
-get(obj_ref::ObjectRef) = get(obj_ref.oid)
-get(oid::ray_jll.ObjectIDAllocated; timeout_ms::Int=-1) = get(ray_jll.get(oid, timeout_ms))
+get(obj_ref::ObjectRef; pollint_seconds=0.1) = get(obj_ref.oid; pollint_seconds)
+function get(oid::ray_jll.ObjectIDAllocated; pollint_seconds=0.1)
+    # we poll here with a 0 timeout in order to provide immediate return and
+    # create a yield point so this is `@async` friendly
+    timeout_ms = 0
+    result = get(ray_jll.get(oid, timeout_ms))
+    while result === nothing
+        sleep(pollint_seconds)
+        result = get(ray_jll.get(oid, timeout_ms))
+    end
+    return result
+end
 get(ray_obj::SharedPtr{ray_jll.RayObject}) = _get(take!(ray_jll.GetData(ray_obj[])))
-get(x) = x
+get(x; kwargs...) = x
 
 function _get(bytes)
     result = deserialize_from_bytes(bytes)
--- a/Ray.jl/src/object_store.jl
+++ b/Ray.jl/src/object_store.jl
@@ -25,16 +25,13 @@ captured exception will be thrown on `get`.
 """
 get(obj_ref::ObjectRef; pollint_seconds=0.1) = get(obj_ref.oid; pollint_seconds)
 
-function get(oid::ray_jll.ObjectIDAllocated; pollint_seconds=0.1)
-    # we poll here with a 0 timeout in order to provide immediate return and
-    # create a yield point so this is `@async` friendly
-    timeout_ms = 0
-    result = get(ray_jll.get(oid, timeout_ms))
-    while result === nothing
-        sleep(pollint_seconds)
-        result = get(ray_jll.get(oid, timeout_ms))
+function get(oid::ray_jll.ObjectIDAllocated)
+    local ray_obj
+    while true
+        ray_obj = ray_jll.get(oid, 0)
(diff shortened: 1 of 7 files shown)


Ray.jl/src/object_store.jl
```diff
@@ -20,7 +20,7 @@ if run in an `@async` task.
 If the task that generated the `ObjectID` failed with a Julia exception, the
 captured exception will be thrown on `get`.
 """
-get(oid::ray_jll.ObjectIDAllocated) = _get(take!(ray_jll.get(oid)))
+get(oid::ray_jll.ObjectIDAllocated; timeout_ms::Int=-1) = _get(take!(ray_jll.get(oid, timeout_ms)))
```
omus, line 23: We need to allow passing in this keyword for each `get` method otherwise users will be unable to write generic functions. 

I'll also mention we're deviating from the Python interface which accepts the timeout in seconds: https://github.com/ray-project/ray/blob/10c461c6d03b11bfa3f03a1a95907b6db27aac7d/python/ray/_private/worker.py#L754. I can understand why just pass in milliseconds as this avoids any floating point issues but since Julia functions typically take seconds for timeouts it is probably best to use seconds here again.

Overall, for the Julia interface I think it would be better to have an `isready` and blocking `get`.

omus, line 23: Having `take!` return a `nothing` and `deserialize_from_bytes(::Nothing)` seems off. What about:
```suggestion
function get(oid::ray_jll.ObjectIDAllocated; timeout_ms::Int=-1)
    buffer = ray_jll.get(oid, timeout_ms)
    return isnull(buffer[]) ? nothing : _get(take!(buffer))
end
```

kleinschmidt: I agree that separating isready and get/fetch is reasonable, but IMO the current situation is not acceptable: right now, `get` on an un-ready future can hard lock the julia session in such a way that the only option is either to kill it at teh OS level or stop the ray server.

kleinschmidt: that is, the blocking is not the issue _per se_, rather that teh blocking happens at a level that we have no control over on teh julia side.  even in an `@async` block, for instance, it'll block indefinitely

omus: Having `ray_jll.get` return a `nullptr` is sensible if we hit the timeout. The interface for `take!` however is to read a `Buffer` and return the data as bytes. I could see having the `isnull` check inside `take!` if we could have the function return a `UInt8[]` but as a non-null `Buffer` could store an empty vector (note any serialized data will be non-empty due to the serialization header) we shouldn't do this.

The only way we'd see segfaults is if we pass in a `nullptr` to `take!`. As these low level functions shouldn't be used by end users I'm not seeing an issue with checking for `nullptr` after the `ray_jll.get` call.

omus: So is the plan to utilize a small internal timeout here until we reach the user's specified timeout? 

What I want to avoid is providing a high-level user interface which we don't want to support in the long term. Did you know about the [`Contains`](https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/core_worker/core_worker.h#L683-L691) core worker function? We could possibly utilize that function to perform a similar kind of check on the Julia side (this is probably how we'd implement `isready`). I'll note I haven't had to use this function myself so we'll have to validate its behavior.

kleinschmidt: > So is the plan to utilize a small internal timeout here until we reach the user's specified timeout?

basically yes, unless we can find some way to do the long-poll get on a separate thread or something.

> What I want to avoid is providing a high-level user interface which we don't want to support in the long term. 

Yup, we're on the same page there.  I think we can have a separation between the `ray_jll.ObjectID` and `Ray.ObjectRef` methods: the objectrefs won't have a manual timeout but the ObjectIDs would...

> Did you know about the [Contains](https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/core_worker/core_worker.h#L683-L691) core worker function? We could possibly utilize that function to perform a similar kind of check on the Julia side (this is probably how we'd implement isready). I'll note I haven't had to use this function myself so we'll have to validate its behavior.

yeah I was just looking at contains, I think that's what we'll want to use.  Will probably want to use a v. short timeout there too to make things async friendly, and so will still need to find some way to handle timeouts.

Ray.jl/src/object_store.jl
```diff
@@ -23,10 +23,29 @@ if run in an `@async` task.
...
+    # we poll here with a 0 timeout in order to provide immediate return and
+    # create a yield point so this is `@async` friendly
+    timeout_ms = 0
+    result = get(ray_jll.get(oid, timeout_ms))
+    while result === nothing
+        sleep(pollint_seconds)
+        result = get(ray_jll.get(oid, timeout_ms))
+    end
+    return result
+end
```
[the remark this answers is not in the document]
kleinschmidt: that's a better idea!

Ray.jl/src/object_store.jl
```diff
@@ -23,10 +23,19 @@ if run in an `@async` task.
 If the task that generated the `ObjectID` failed with a Julia exception, the
 captured exception will be thrown on `get`.
 """
-get(obj_ref::ObjectRef) = get(obj_ref.oid)
-get(oid::ray_jll.ObjectIDAllocated) = get(ray_jll.get(oid))
+get(obj_ref::ObjectRef; pollint_seconds=0.1) = get(obj_ref.oid; pollint_seconds)
```
omus:
```suggestion
get(obj_ref::ObjectRef) = get(obj_ref.oid)
```

Ray.jl/src/object_store.jl
```diff
@@ -23,10 +23,19 @@ if run in an `@async` task.
...
+    while true
+        ray_obj = ray_jll.get(oid, 0)
+        isnull(ray_obj[]) ? sleep(0.1) : break
+    end
+    return get(ray_obj)
+end
+
 get(ray_obj::SharedPtr{ray_jll.RayObject}) = _get(take!(ray_jll.GetData(ray_obj[])))
-get(x) = x
+get(x; kwargs...) = x
```
omus:
```suggestion
get(x) = x
```

src/wrappers/any.jl
```diff
@@ -157,6 +157,7 @@ end
 #####
 
 function Base.take!(buffer::CxxWrap.CxxWrapCore.SmartPointer{<:Buffer})
+    isnull(buffer[]) && return nothing
```
omus, line 160:
```suggestion
```

Ray.jl/src/object_store.jl
kleinschmidt: ha, jinx ;)

omus (approved):
Beauty! Thanks for working through this

Issue #77, opened 2023-09-01
Support passing object references between workers


omus:
At the moment when an object reference is created it must be used on the worker where it was created. For example:

```julia
julia> using Ray

julia> Ray.init()

julia> x = Ray.put(1)
ray_core_worker_julia_jll.ObjectIDAllocated(Ptr{Nothing} @0x0000600003a15140)

julia> y = Ray.submit_task(identity, (x,))
ray_core_worker_julia_jll.ObjectIDAllocated(Ptr{Nothing} @0x000060000392f990)

julia> Ray.get(y)
ray_core_worker_julia_jll.ObjectIDAllocated(Ptr{Nothing} @0x0000000000000000)

julia> Ray.get(Ray.get(y))
ERROR: C++ object of type N3ray8ObjectIDE was deleted
Stacktrace:
 [1] get(arg1::Union{ConstCxxRef{ray_core_worker_julia_jll.ObjectID}, CxxRef{ray_core_worker_julia_jll.ObjectID}, ray_core_worker_julia_jll.ObjectID})
   @ ray_core_worker_julia_jll ~/.julia/packages/CxxWrap/aXNBY/src/CxxWrap.jl:624
 [2] get(oid::ray_core_worker_julia_jll.ObjectIDAllocated)
   @ Ray ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/object_store.jl:23
 [3] top-level scope
   @ REPL[5]:1
```

There are two issues going on here:
1. [Julia serializes `Ptr`s as `C_NULL`](https://github.com/JuliaLang/julia/blob/ec2f1d3c0b2a37b5acafeaab2454fedbd83a31ab/stdlib/Serialization/src/Serialization.jl#L190)
2. Ray needs to be informed about object references contained within serialized data (see Ray's  [`prepare_args_internal` Cython code](https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/python/ray/_raylet.pyx#L743-L746))

Implementing 1 without 2 results in:
```julia
julia> Ray.get(obj_ref)
[...]

PR #67, opened 2023-08-29
Support returning `ObjectRef` from tasks

omus:
Supports serialization of `ObjectID` via the wrapper type `ObjectRef` to address this problem:
```julia
julia> using Ray, Serialization

julia> Ray.init()

julia> ref = Ray.put(1)
ray_core_worker_julia_jll.ObjectIDAllocated(iPtr{Nothing} @0x0000600003899080)

julia> ref2 = deserialize(IOBuffer(sprint(serialize, ref)))
ray_core_worker_julia_jll.ObjectIDAllocated(Ptr{Nothing} @0x0000000000000000)

julia> Ray.get(ref2)
ERROR: C++ object of type N3ray8ObjectIDE was deleted
Stacktrace:
 [1] get(arg1::Union{CxxWrap.CxxWrapCore.ConstCxxRef{ray_core_worker_julia_jll.ObjectID}, CxxWrap.CxxWrapCore.CxxRef{ray_core_worker_julia_jll.ObjectID}, ray_core_worker_julia_jll.ObjectID})
   @ ray_core_worker_julia_jll ~/.julia/packages/CxxWrap/aXNBY/src/CxxWrap.jl:624
 [2] get(oid::ray_core_worker_julia_jll.ObjectIDAllocated)
   @ Ray ~/.julia/dev/ray_core_worker_julia_jll/Ray.jl/src/object_store.jl:29
 [3] top-level scope
   @ REPL[7]:1

julia> Ray.get(ref)
1
```

We can probably also get this working without introducing `ObjectRef` but I kept wanting to add Julian functionality to `ray_jll.ObjectIDAllocated` and it just made more sense to have a Julia type which handled that in Ray.jl instead.

PR #81, opened 2023-09-05
Introduce `ObjectRef`

omus:
Introduces the `ObjectRef` Julia object which is the Julian interface for dealing with object references. This is similar to how Python handles things. Having this struct allows us to hide the internals of using C++ `ObjectID`s from users.

Addresses part of #77. Extracted from #67

### DIFF
--- a/Ray.jl/src/object_ref.jl
+++ b/Ray.jl/src/object_ref.jl
@@ -1,0 +1,14 @@
+struct ObjectRef
+    oid::ray_jll.ObjectIDAllocated
+end
+
+ObjectRef(hex_str::AbstractString) = ObjectRef(ray_jll.FromHex(ray_jll.ObjectID, hex_str))
+Base.string(obj_ref::ObjectRef) = String(ray_jll.Hex(obj_ref.oid))
+Base.:(==)(a::ObjectRef, b::ObjectRef) = string(a) == string(b)
+
+function Base.show(io::IO, obj_ref::ObjectRef)
+    write(io, "ObjectRef(\"", string(obj_ref), "\")")
+    return nothing
+end
+
+Base.convert(::Type{ray_jll.ObjectID}, obj_ref::ObjectRef) = obj_ref.oid
--- a/Ray.jl/src/object_ref.jl
+++ b/Ray.jl/src/object_ref.jl
@@ -10,5 +10,3 @@ function Base.show(io::IO, obj_ref::ObjectRef)
     write(io, "ObjectRef(\"", hex_identifier(obj_ref), "\")")
     return nothing
 end
-
-Base.convert(::Type{ray_jll.ObjectID}, obj_ref::ObjectRef) = obj_ref.oid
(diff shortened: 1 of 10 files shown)


Ray.jl/test/object_ref.jl
```diff
@@ -0,0 +1,20 @@
+@testset "ObjectRef" begin
+    @testset "basic" begin
+        hex_str = "f" ^ (2 * 28)
+        obj_ref = ObjectRef(hex_str)
+        @test obj_ref == ObjectRef(hex_str)
+        @test string(obj_ref) == hex_str
+    end
+
+    @testset "show" begin
+        hex_str = "f" ^ (2 * 28)
```
kleinschmidt, line 10: F

Ray.jl/src/object_store.jl
```diff
@@ -20,6 +22,7 @@ if run in an `@async` task.
 If the task that generated the `ObjectID` failed with a Julia exception, the
 captured exception will be thrown on `get`.
 """
+get(obj_ref::ObjectRef) = _get(take!(ray_jll.get(convert(ray_jll.ObjectID, obj_ref.oid))))
```
kleinschmidt, line 25: what's the point of the `convert` in there?  should be an `ObjectIDAllocated <: ObjectID` already right?

Ray.jl/src/object_ref.jl
```diff
@@ -0,0 +1,14 @@
+struct ObjectRef
+    oid::ray_jll.ObjectIDAllocated
+end
+
+ObjectRef(hex_str::AbstractString) = ObjectRef(ray_jll.FromHex(ray_jll.ObjectID, hex_str))
+Base.string(obj_ref::ObjectRef) = String(ray_jll.Hex(obj_ref.oid))
```
kleinschmidt, line 6: I think this should be something else; the julia docs say that `Base.string` and `Base.show` should be consistent, and they aren't in this case.

kleinschmidt: ah actually I misread, its' _print_ that should be consistent with `string`.  I still dont' think we want to be `print`ing these as a string of hex digits without any additional disambiguation though

Ray.jl/src/object_store.jl
omus: Whoops, I made a mistake here. This was intended to be:
```suggestion
get(obj_ref::ObjectRef) = _get(take!(ray_jll.get(convert(ray_jll.ObjectID, obj_ref))))
```
Felt a bit odd doing the field access here but either should work. What's your preference?

kleinschmidt: I think field access is fine!  This is defining the translation layer so it's fine IMO to access the internal fields, right?

kleinschmidt: the `_get(take!(get` turducken is a little funny but I understand the reason for the different levels of indirection

Ray.jl/src/object_ref.jl
omus: I'll create an accessor called `hex_identifier` instead

kleinschmidt (approved):

PR #83, opened 2023-09-05
C++ wrapped `put`/`get` work with `RayObject`

omus:
Part of the work for #77. In order to specify object IDs contained within an object I needed to update our `put`/`get` wrapped functions. By updating these functions to pass in a `RayObject` these are more inline with the `ray::core::CoreWorker` interface and also allow me to specify the `nested_refs` argument for `RayObject`.

### DIFF
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -183,12 +183,15 @@ ray::core::CoreWorker &_GetCoreWorker() {
 }
 
 // Example of using `CoreWorker::Put`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L224-L237
-void put(const std::shared_ptr<RayObject> object,
-         const std::vector<ObjectID> &contained_object_ids,
-         ObjectID *object_id) {
+ObjectID put(const std::shared_ptr<RayObject> object,
+             const std::vector<ObjectID> &contained_object_ids) {
 
     auto &worker = CoreWorkerProcess::GetCoreWorker();
-    RAY_CHECK_OK(worker.Put(*object, contained_object_ids, object_id));
+
+    // Store our data in the object store
+    ObjectID object_id;
+    RAY_CHECK_OK(worker.Put(*object, contained_object_ids, &object_id));
+    return object_id;
 }
 
 // Example of using `CoreWorker::Get`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L210-L220
--- a/deps/wrapper.cc
+++ b/deps/wrapper.cc
@@ -183,16 +183,16 @@ ray::core::CoreWorker &_GetCoreWorker() {
 }
 
 // https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L224-L237
-void put(std::shared_ptr<RayObject> ray_obj,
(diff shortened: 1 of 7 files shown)


deps/wrapper.cc
```diff
@@ -182,33 +182,29 @@ ray::core::CoreWorker &_GetCoreWorker() {
...
+// Example of using `CoreWorker::Get`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L210-L220
+std::shared_ptr<RayObject> get(const ObjectID object_id) {
     auto &worker = CoreWorkerProcess::GetCoreWorker();
 
     // Retrieve our data from the object store
-    std::vector<std::shared_ptr<RayObject>> results;
+    std::vector<std::shared_ptr<RayObject>> objects;
     std::vector<ObjectID> get_obj_ids = {object_id};
-    RAY_CHECK_OK(worker.Get(get_obj_ids, -1, &results));
+    RAY_CHECK_OK(worker.Get(get_obj_ids, -1, &objects));
```
kleinschmidt, line 204: kind of orthogonal to this PR but we shouldn't be using `RAY_CHECK*` because when it fails, it exits the process.  better to check + throw.  but we can handle that in a cleanup later.

deps/wrapper.cc
```diff
@@ -182,33 +182,29 @@ ray::core::CoreWorker &_GetCoreWorker() {
     return CoreWorkerProcess::GetCoreWorker();
 }
 
-// https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L224-L237
-ObjectID put(std::shared_ptr<Buffer> buffer) {
+// Example of using `CoreWorker::Put`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L224-L237
+ObjectID put(const std::shared_ptr<RayObject> object,
+             const std::vector<ObjectID> &contained_object_ids) {
```
kleinschmidt, line 187: aren't the contained object IDs a field in the ray object itself?  I guess the core worker Put function does take this separately: https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/core_worker/core_worker.cc#L1110-L1112

kleinschmidt (approved):
looks good.  one comment to follow up on later but doesn't need to be handled here.

deps/wrapper.cc
omus: I also find this strange. I'll need to do some experimentation with this yet to determine why both are needed but for this PR in particular it seemed good to provide the option for experimentation.

kleinschmidt: yup, no reason not to match the interface from the c++ function here

deps/wrapper.cc
omus: I can make a follow up issue for this. The `CoreWorker::Get` usually returns the `Status` and lets the caller decide what to do. We should probably do the same but as you said that can be done as a follow up.

omus: https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/issues/84

kleinschmidt: yeah, this is an issue for #78 as well.  I think potentially we could return a tuple or a wrapper of some kind.

PR #97, opened 2023-09-07
Return a `RayObject` from `task_executor`

omus:
Allows us to specify any inlined object references in the return.


omus:
As far as codecov is concerned `task_executor` isn't tested. If we enable `--code-coverage=true` on the worker processes we can get this for free.

glennmoy (approved):

PR #98, opened 2023-09-07
wrap `CoreWorker::Contains`

kleinschmidt:
This will provide support for `wait`/`isready`/`get` polling without data movement.

There's an additional output that I'm not capturing (whether something is local vs. in plasma store); I don't think that it's really necessary and we can always add a flag or another function to check for only local or only plasma later if need be.

### DIFF
--- a/ray_julia_jll/test/function_descriptor.jl
+++ b/ray_julia_jll/test/function_descriptor.jl
@@ -1,3 +1,5 @@
+using ray_julia_jll: JuliaFunctionDescriptor, function_descriptor
+
 @testset "function descriptor" begin
     fd = function_descriptor(isless)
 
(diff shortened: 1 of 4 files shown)


glennmoy (approved):

ray_julia_jll/test/function_descriptor.jl
```diff
@@ -1,3 +1,5 @@
+using ray_julia_jll: JuliaFunctionDescriptor, function_descriptor
```
omus, line 1: Should we move this to `runtests.jl`?

omus (approved):

ray_julia_jll/test/function_descriptor.jl
kleinschmidt: that's where it was, but other test files have their own `using`s and it seemed weird to have ONLY these imports at the top level

Issue #86, opened 2023-09-06
Roadmap for a `Future`-like interface to `ObjectRef`


kleinschmidt:
Distributed.jl `Future` provides a reasonably well-known interface for remote, async return values.  Like ObjectRefs, they are write-once containers.  It's probably good to try to provide a consistent interface for users' interactions with `ObjectRef`.

Here's what Future provides:
- `Base.fetch` - wait for and retrieve the value
- `Base.put!` - store a value 
- `Base.isready` - check whether a value is ready to be `fetch`ed
- `Base.wait` - block until a value is ready

I think `fetch` is what we (and Ray) call `get`, `put!` is `put`, and we haven't implemented `isready` and `wait` but those are pretty straightforward (essentially just wrapping `Contains` with a short or 0ms timeout, as in #78 )

PR #99, opened 2023-09-07
use `ray_julia_jll.contains` in `get`, add `Base.isready`/`wait` methods

kleinschmidt:
Work towards #86.

~~Depends on #98~~

### DIFF
--- a/src/object_store.jl
+++ b/src/object_store.jl
@@ -24,7 +24,8 @@ captured exception will be thrown on `get`.
 """
 function get(obj_ref::ObjectRef)
     wait(obj_ref)
-    ray_obj = ray_jll.get(obj_ref.oid, -1)
+    ray_obj = ray_jll.get(obj_ref.oid, 0)
+    isnull(ray_obj[]) && error("got null pointer after successful `wait`; this is a bug!")
     return get(ray_obj)
 end
 
--- a/src/object_store.jl
+++ b/src/object_store.jl
@@ -25,11 +25,10 @@ captured exception will be thrown on `get`.
 get(obj_ref::ObjectRef) = get(obj_ref.oid)
 
 function get(oid::ray_jll.ObjectIDAllocated)
-    local ray_obj
-    while true
-        ray_obj = ray_jll.get(oid, 0)
-        isnull(ray_obj[]) ? sleep(0.1) : break
+    while !ray_jll.contains(oid)
+        sleep(0.1)
     end
+    ray_obj = ray_jll.get(oid, -1)
     return get(ray_obj)
 end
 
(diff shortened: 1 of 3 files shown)


src/object_store.jl
```diff
@@ -25,11 +25,10 @@ captured exception will be thrown on `get`.
 get(obj_ref::ObjectRef) = get(obj_ref.oid)
 
 function get(oid::ray_jll.ObjectIDAllocated)
-    local ray_obj
-    while true
-        ray_obj = ray_jll.get(oid, 0)
-        isnull(ray_obj[]) ? sleep(0.1) : break
+    while !ray_jll.contains(oid)
+        sleep(0.1)
     end
```
glennmoy, line 30: can this now be refactored to use `Base.wait`?

kleinschmidt: ah yes, good catch.

kleinschmidt: ah wait, maybe not.  unless we define `Base.wait` on `ObjectID`

src/object_store.jl
```diff
@@ -22,14 +22,9 @@ is blocking until the data is available in the local object store.
...
-
-function get(oid::ray_jll.ObjectIDAllocated)
-    local ray_obj
-    while true
-        ray_obj = ray_jll.get(oid, 0)
-        isnull(ray_obj[]) ? sleep(0.1) : break
-    end
+function get(obj_ref::ObjectRef)
+    wait(obj_ref)
+    ray_obj = ray_jll.get(obj_ref.oid, -1)
```
omus, line 27: Should we set the timeout to zero just in case something weird happens and we don't block forever?
```suggestion
    ray_obj = ray_jll.get(obj_ref.oid, 0)
```

src/object_store.jl
```diff
@@ -43,3 +38,21 @@ function _get(bytes)
...
+    Base.isready(obj_ref::ObjectRef)
+
+Check whether `obj_ref` has a value that's ready to be retrieved.
+"""
+Base.isready(obj_ref::ObjectRef) = ray_jll.contains(obj_ref.oid)
+
+"""
+    Base.wait(obj_ref::ObjectRef)
+
+Block until `isready(obj_ref)`.  Returns `nothing`.
```
omus, line 51:
```suggestion
    Base.wait(obj_ref::ObjectRef) -> Nothing

Block until `isready(obj_ref)`.
```

test/task.jl
```diff
@@ -203,6 +203,8 @@ end
     # driver is shutdown we'll see the error: `ERROR: Package Ray errored during testing`
     # https://github.com/beacon-biosignals/Ray.jl/issues/96
     wait(t)
+
+    @test isready(obj_ref)
```
omus, line 207: Test that `@test !isready(obj_ref)` before the `wait`

omus (approved):
Minor tweaks

src/object_store.jl
kleinschmidt: I dunno, there _might_ be some kind of weird race condition here that prevents it from actually returning RIGHT away.  I guess we could switch to polling `get` at this point though...or check for a nullptr and throw an error (and if folks encounter it too often, that's a signal for us to fix it...)

PR #102, opened 2023-09-08
Add `has_owner` accessor for `ObjectRef`

omus:
Was found to be useful in determining if an ObjectID has been registered with the core worker. If an `ObjectRef` does not have an owner then a `Ray.get` call will wait indefinitely. See https://github.com/beacon-biosignals/Ray.jl/issues/77#issuecomment-1712129853 for details.


kleinschmidt (commented):
Don't see anything objectionable here but woudl be good to indclude in the PR text a sentence or two about why this is necessary (especially given that the `HasOwner` method is private on the core worker)

kleinschmidt (approved):

PR #110, opened 2023-09-12
Support `readheader` with `RaySerializer`

omus:
While debugging a serialization issue with #77 it was noticed that `readheader` couldn't be called on a `RaySerializer`. This PR adds the `version` field so that this call succeeds.

Without this change a `deserialize(::RaySerializer)` call will fail. We haven't seen this issue before as we've only been using `RaySerializer` for serialization so far.

### DIFF
--- a/src/Ray.jl
+++ b/src/Ray.jl
@@ -14,7 +14,7 @@ using Logging
 using LoggingExtras
 using Pkg
 using Serialization: Serialization, AbstractSerializer, Serializer, deserialize,
-    reset_state, serialize, serialize_type, ser_version, writeheader
+    reset_state, serialize, serialize_type, writeheader
 
 import ray_julia_jll as ray_jll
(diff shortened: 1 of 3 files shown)


src/Ray.jl
```diff
@@ -14,7 +14,7 @@ using Logging
 using LoggingExtras
 using Pkg
 using Serialization: Serialization, AbstractSerializer, Serializer, deserialize,
-    reset_state, serialize, serialize_type, writeheader
+    reset_state, serialize, serialize_type, ser_version, writeheader
```
kleinschmidt, line 17: name is cryptic enough that I think it helps to preface with `Serialization` wherever it is used...

```suggestion
    reset_state, serialize, serialize_type, writeheader
```

kleinschmidt (approved):

PR #111, opened 2023-09-12
Expanded setup instructions

kleinschmidt:
This is based on feedback from early internal Beacon users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,5 +94,5 @@ cd -
 
 ```sh
 # Run the tests
-julia --project -e 'using Pkg; Pkg.test()'
+julia --project -e 'using Pkg; Pkg.develop(; path="./ray_julia_jll"); Pkg.test()'
 ```
--- a/README.md
+++ b/README.md
@@ -4,31 +4,95 @@ The `ray_julia_jll` is the Julia C++ wrapper that interfaces with the Ray.io pro
 
 At the moment the package requires that the shared libraries are built directly on the host system and does not provide any precreated cross-compiled libraries. We hope to turn this package into a full fledged JLL but this direct build process will work as we experiment with interfacing with the Ray core worker C++ interface.
 
-## Example
+## Setup (manual/dev build)
+
+Ultimately our aim is to make installing Ray.jl as simple as `Pkg.add("Ray")`.  However, for the time being, it's still necessary to install the full build toolchain required to build Ray _and_ our Ray.jl wrappers.
+
+Start by cloning this repo.  The following instructions assume you start from the repository root as the working directory:
 
 ```sh
-# We recommend always using the same virtualenv as Bazel can cause unnecessary rebuilds when
-# using different versions of Python.
+git clone https://github.com/beacon-biosignals/Ray.jl.git
+cd Ray.jl
+```
+
+### Install dev dependnecies
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,31 +4,95 @@ The `ray_julia_jll` is the Julia C++ wrapper that interfaces with the Ray.io pro
...
+
+Alternatively, if you have `npm` installed you can install Bazelisk that way:
+
+```sh
+npm install -g @bazel/bazelisk
+```
+
+### Prepare python environment
+
+We recommend always using the same virtualenv as Bazel can cause unnecessary rebuilds when using different versions of Python.
```
glennmoy, line 61:
```suggestion
We recommend always using the same virtualenv since Bazel can cause unnecessary rebuilds when using different versions of Python.
```

l e g i b i l i t y

README.md
```diff
@@ -4,31 +4,95 @@ The `ray_julia_jll` is the Julia C++ wrapper that interfaces with the Ray.io pro
...
-## Sources
-
-The tarballs for `ray_julia_jll.jl` have been built from these sources:
+### Validate installation
 
-* ray v2.5.1
-* libcxxwrap_julia
+```sh
+# Run the tests
+julia --project -e 'using Pkg; Pkg.develop(; path="./ray_julia_jll"); Pkg.test()'
```
glennmoy, line 109: is a side-effect of this to get `Ray.jl` to `dev` `ray_julia_jll`?

If so might be good to call that out? otherwise I'd wonder why we don't just do `--project=ray_julia_jll`

README.md
```diff
@@ -4,31 +4,95 @@ The `ray_julia_jll` is the Julia C++ wrapper that interfaces with the Ray.io pro
...
-# We recommend always using the same virtualenv as Bazel can cause unnecessary rebuilds when
-# using different versions of Python.
 python3 -m venv venv
 source venv/bin/activate
 
 # Build the required libraries
 julia --project=ray_julia_jll -e 'using Pkg; Pkg.build(verbose=true)'
+```
+
+### Build Ray CLI/server
```
glennmoy, line 95:
```suggestion
From within the same virtual environment you can also easily install the ray cli by running the command below. First, modify the url below to correspond to the version of python you are using (in this instance it's `python3.8`).

```sh
pip install https://github.com/beacon-biosignals/ray/releases/download/ray-2.5.1-beacon/ray-2.5.1-cp38-cp38-manylinux2014_x86_64.whl
```

On Mac it's:
```sh
pip install https://github.com/beacon-biosignals/ray/releases/download/ray-2.5.1-beacon/ray-2.5.1-cp38-cp38-macosx_13_0_arm64.whl
```

Alternatively, to build ray from source (which takes ~30 mins) you can follow these steps:
```

glennmoy: the suggestion is getting messed up by the ```suggestion` block but you get the idea

glennmoy (approved):
I assume you had @dhanyscode run through this so I'll take for granted that the commands and dependencies all check out

jmmanley (approved):
These install instructions worked for me (Ray tests passed) on MacOS ARM / Python 3.10 using the provided github release!🎉

README.md
kleinschmidt: The Ray.jl tests are at this point a bit more extensive.  Dev'ing the jll is necessary since otherwise it tries to `add` all the dependencies when instantiating the test environment; if there's a manifest alreayd present it'll re-use taht to the extent that it can.

I do think it's worth calling it out (that you currently need to `dev` _both_ the Ray.jl and jll)

README.md
kleinschmidt: I thought I had these but maybe we were racin'

PR #115, opened 2023-09-13
Utils to build and push JLL artifacts to GH releases page

glennmoy:
Templated off https://github.com/JuliaTime/TZJData.jl/blob/main/gen/make.jl

From the playbook in https://github.com/beacon-biosignals/Ray.jl/issues/124 I'm breaking this up into two stages of work
1. (This PR) Run make.jl on each platform/version to create the initial tag and draft pre-release, and to generate+upload the tarball assets.
2. (followup: https://github.com/beacon-biosignals/Ray.jl/pull/134) Run artifacts.jl _once_ to update the Artifacts.toml and push commits to branch. Will also include the CI to check the artifacts.

Release created from the script: https://github.com/beacon-biosignals/Ray.jl/releases/tag/v0.1.0

Final steps after PR is merged will be manual: tag switcheroo and register tag.

Note: I've opted to remove the `draft` element of the releases since every time we upload an asset it creates a new draft instead of appending the existing one. Using `replace` etc has no effect.

TODO
- [x] When building a JLL check that...
    - [x] ...only known list of items are baked into tarball
    - [x] ...running on the same platform but different julia version updates release but no new tag
    - [x] ...running on a different platform updates release but no new tag
    - [x] ...running on the same platform / same julia version fails
- [x] Add README for running `make.jl`

---
  - https://app.asana.com/0/0/1205455951987952


ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,171 @@
...
+        $tag $path
+    ```
+
+    run(cmd)
+end
+
+if abspath(PROGRAM_FILE) == @__FILE__
+
+    repo_path = abspath(joinpath(@__DIR__, "..", ".."))
+    pkg_url = remote_url(repo_path)
```
glennmoy, line 113: `pkg_url` I'm getting back is `"git@github.com:beacon-biosignals/ray.jl"` which is failing the checks in `upload_to_github_release`

The full `artifact_url` is `"git@github.com:beacon-biosignals/ray.jl/releases/download/v0.1.0/ray_julia.v0.1.0.aarch64-apple-darwin-libgfortran5-cxx11-julia_version+1.9.2.tar.gz"` which, when parsed as a URI has `achive_uri.host == ""`.

Secondly the `archive_ur.path` doesn't have the leading `/` so the regex check is also failing.

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,171 @@
...
+        artifacts_toml,
+        "ray_julia_jll",
+        tree_hash_sha1(tarball_path);
+        download_info=[(artifact_url, sha256sum(tarball_path))],
+        force=true
+    )
+
+    # TODO: Ensure no other files are staged before committing
+    @info "Committing and pushing changes to Artifacts.toml"
+    branch = "gm/test"  # TODO - pass as function arg
```
glennmoy, line 148: Since this won't be pusing to`main` the branch name will probably have to be passed in as an arg to the script.

Moreover the branch will have to be repeatedly `commited` to if we want to bundle each platform/julia version into the same PR

ray_julia_jll/gen/make.jl
omus: You probably need to be using an HTTPS URL in order for `upload_to_github_release` to work instead of this SCP-style syntax. I would recommend to just convert the SCP-style syntax to an HTTP URL. There's a regex I wrote for the LibGit2 library which you can use to parse both: https://github.com/JuliaLang/julia/blob/572fa5055ecb709edd3d227a10b46549f82b72b2/stdlib/LibGit2/src/utils.jl#L9C7-L9C16

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,171 @@
...
+    jll_version = jll_project.version
+
+    # e.g. ray_julia.v0.1.0.aarch64-apple-darwin-libgfortran5-cxx11-julia_version+1.9.2
+    host_platform = BinaryPlatforms.host_triplet()
+    tarball_name = "ray_julia.v$jll_version.$host_platform.tar.gz"
+
+    # Build JLL
+    # TODO: execute inside a python venv
+    Pkg.build("ray_julia_jll"; verbose=true)
+    compiled_dir = joinpath(repo_path, "ray_julia_jll", "deps", "bazel-bin")
```
omus, line 127: May want to add a TODO on limiting what content here we include in the tarball

omus (commented):
Partial review

ray_julia_jll/gen/make.jl
glennmoy: I've opted to push the Artifacts.toml changes to the currently checked out branch. 
A few things we should check in this regard:

- the branch is clean (no unstaged commits)
- the Artifacts.toml is the only change staged/committed
- probably something about it being a new branch...

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,168 @@
...
+using URIs: URI
+using ghr_jll: ghr
+
+const ASSETS = Set(["external",
+                    "julia_core_worker_lib.so-2.params",
+                    "_objs",
+                    "julia_core_worker_lib.so.runfiles_manifest",
+                    "julia_core_worker_lib.so",
+                    "julia_core_worker_lib.so.cppmap",
+                    "julia_core_worker_lib.so.runfiles"])
```
omus, line 16: Do we need anything more than the `.so` file?

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,168 @@
...
+
+    # Check for this now before we spend any time building the JLL
+    !haskey(ENV, "GITHUB_TOKEN") && error("\"GITHUB_TOKEN\" environment variable required.")
+
+    repo_path = abspath(joinpath(@__DIR__, "..", ".."))
+    pkg_url = remote_url(repo_path)
+
+    # Build JLL
+    @info "Building ray_julia_jll..."
+    Pkg.build("ray_julia_jll"; verbose=true)
```
omus, line 136: Once artifacts are the norm for this package we won't use a `deps/build.jl` script anymore. That code will still need to exist but `Pkg.build("ray_julia_jll")` shouldn't do anything.

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,168 @@
...
+    tarball_name = "ray_julia.v$jll_version.$host_triplet.tar.gz"
+
+    @info "Creating tarball $tarball_name"
+    tarball_path = joinpath(tempdir(), tarball_name)
+    create_tarball(readlink(compiled_dir), tarball_path)
+
+    m = match(URL_REGEX, pkg_url)
+    if m === nothing
+        throw(ArgumentError("Package URL is not a valid SCP or HTTP URL: $(pkg_url)"))
+    end
```
omus, line 160: Consider making this into a function `parse_git_remote_url(::AbstractString) -> URI`

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,168 @@
...
+
+    pkg_http_url = "https://" * joinpath(m[:host], m[:path])
+    tag = "v$(jll_version)"
+    artifact_url = "$(pkg_http_url)/releases/download/$(tag)/$(basename(tarball_path))"
+
+    branch = LibGit2.with(LibGit2.branch, LibGit2.GitRepo(repo_path))
```
omus, line 166:
```suggestion
    @info "Uploading $(basename(tarball_path) to $artifact_url"
    branch = LibGit2.with(LibGit2.branch, LibGit2.GitRepo(repo_path))
```

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,168 @@
+using Base: BinaryPlatforms
+using CodecZlib: GzipCompressorStream, GzipDecompressorStream
```
omus, line 2:
```suggestion
using CodecZlib: GzipCompressorStream
```

ray_julia_jll/gen/make.jl
omus: May want to break this up into a `gen/build.jl` and a `gen/upload.jl`

ray_julia_jll/gen/make.jl
glennmoy: perhaps not...but we'd have to copy the files out into another tmpdir and then create the tarball from that source.

ray_julia_jll/gen/make.jl
glennmoy: Hmm...if we break it up then we'd have to pass `clean=false` to `mktempdir` or (probably better) create a local directory with a given name that we add to `.gitignore`.

glennmoy: re `Pkg.build` I think I'll deprecate that as part of https://github.com/beacon-biosignals/Ray.jl/pull/134 since  the package won't be able to pull the JLLs until we've updated the Artifacts anyway...

omus: Works for me

ray_julia_jll/gen/README.md
```diff
@@ -0,0 +1,17 @@
+# Building and Publishing JLLs
+
+1. Run the `make.jl` script to build the tarballs for the host platform and julia-version. 
+This will create a GitHub pre-release for the tag in the `Project.toml`.
+Note: It is advised you run this within the python virtual environment associated with the root package and set a `GITHUB_TOKEN` environment variable.
```
omus, line 5: Mention what platforms are required here

ray_julia_jll/gen/make.jl
```diff
@@ -0,0 +1,170 @@
...
+using Pkg.Types: read_project
+using Tar: Tar
+using URIs: URI
+using ghr_jll: ghr
+
+const SO_FILES = ["julia_core_worker_lib.so-2.params",
+                 "julia_core_worker_lib.so.runfiles_manifest",
+                 "julia_core_worker_lib.so",
+                 "julia_core_worker_lib.so.cppmap",
+                 "julia_core_worker_lib.so.runfiles"]
```
omus, line 14: Just checking if these are all needed or just "julia_core_worker_lib.so"

omus (approved):
Some minor final suggestions

ray_julia_jll/gen/make.jl
glennmoy: done

ray_julia_jll/gen/README.md
```diff
@@ -0,0 +1,25 @@
+# Building and Publishing JLLs
+
+1. Run the `make.jl` script to build the tarball for the host platform and a given julia-version.
+Builds are required for `linux x86_64` and `macos-arm64` on julia `v1.8.0` and `v1.9.0`.
```
omus, line 4: I believe you can use the latest version of this minor releases to make the builds?
```suggestion
Builds are required for `linux x86_64` and `macos-arm64` on julia `v1.8` and `v1.9`.
```

ray_julia_jll/gen/README.md
```diff
@@ -0,0 +1,25 @@
...
+It is advised you run this within the python virtual environment associated with the root package.
+Note: rerunning this script _will overwrite_ an existing tarball.
+```
+julia --project -e 'using Pkg; Pkg.develop(path=".."); Pkg.instantiate()'
```
omus, line 8: I think for the `gen` project it would be okay to check in the `Manifest.toml` which would allow you to skip the `Pkg.develop(path="..")` instruction. Doesn't matter much though.

ray_julia_jll/gen/README.md
```diff
@@ -0,0 +1,25 @@
...
+julia --project make.jl
+```
+
+2. Once all tarballs have been built for a given host you can publish them as assets to a GitHub pre-release by running the `upload.jl` script, which requires a `GITHUB_TOKEN` environment variable.
+Note: reruning this script will only upload new tarballs and skip any that have already been published.
```
omus, line 13: Is this still true? Now you can run `upload.jl` after each tarball has been built and you don't need to wait for all of them to complete. In fact this is possibly misleading as you need to run `upload.jl` for each tarball now.

ray_julia_jll/gen/README.md
glennmoy: I wasn't sure... in JuliaBinaryWrappers I never saw anything other than `1.X.0` being used...

e.g. https://github.com/JuliaBinaryWrappers/Geant4_julia_jll.jl/tree/main/src/wrappers

ray_julia_jll/gen/README.md
glennmoy: You can either
a) run `upload.jl` once all tarballs have been built and (assuming it's a fresh release) all will be uploaded
b) run `upload.jl` after each tarball is built in which case it'll only upload the new one

`ghr` doesn't overwrite existing assets by default - I kept this in place as overwritten assets would be irrecoverable

ray_julia_jll/gen/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+export GITHUB_TOKEN="<github_token>"
+julia --project upload.jl
+```
+
+3. Once all assets are uploaded, create a PR which updates the Artifacts.toml with the corresponding artifact URLs.
```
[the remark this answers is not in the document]
glennmoy: that is going to be in https://github.com/beacon-biosignals/Ray.jl/pull/134

ray_julia_jll/gen/upload.jl
```diff
@@ -0,0 +1,121 @@
...
+using ghr_jll: ghr
+
+include("common.jl")
+
+const TARBALL_REGEX = r"""
+    ^ray_julia.v(?<jll_version>([0-9]\.){3})
+    (?<triplet>[a-z, 0-9, \-, \_,]+)
+    -julia_version\+(?<julia_version>([0-9]\.){3})
+    tar.gz$
+    """x
```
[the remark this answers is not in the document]
glennmoy: yep you did that in the follow up PR https://github.com/beacon-biosignals/Ray.jl/pull/134#discussion_r1333152707

I'll just backport those changes here

ray_julia_jll/gen/README.md
omus: Okay, We should call this out explicitly then. It's easy to miss

ray_julia_jll/gen/README.md
omus: Alright. Based upon these instructions there's an implicit step of collecting all the the tarballs to one machine

ray_julia_jll/gen/upload.jl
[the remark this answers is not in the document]
omus: Ah, okay. I thought I was losing my mind. The suggestion here should be more accurate

ray_julia_jll/gen/README.md
glennmoy: Not necessarily...you can switch machines and call `make`/`upload` and the new assets will just append the existing release.
For instance, I built and uploaded the two macos assets, then switched machines and repeated the same commands for the linux x86_64 without needing to copy over the macos tarballs.

ray_julia_jll/gen/README.md
glennmoy: So I did a quick check on this and on julia 1.9 I was able to download and use the 1.9.2 asset (and run tests) so I think we're actually ok to only specify the minor version

glennmoy: kept it as `1.8` and `1.9`

ray_julia_jll/gen/README.md
glennmoy: done

ray_julia_jll/gen/README.md
glennmoy: updated the README to indicate you can run once _or_ after each build

PR #117, opened 2023-09-14
Remove unused `start_worker` argument to default_worker

omus:
Minor cleanup. Depends on PR: https://github.com/beacon-biosignals/ray/pull/9.

Note that if https://github.com/beacon-biosignals/ray/pull/9 is merged prior to this PR being merged then `start_worker()` will break as we currently require this unused argument.


omus:
Turns out that if the backend still passes in the argument then our tests here will hang as the workers never come up due to:
```
too many arguments
usage: <PROGRAM> [--ray_raylet_socket_name RAYLET_SOCKET]
                 [--ray_plasma_store_socket_name STORE_SOCKET]
                 [--ray_address ADDRESS]
                 [--ray_node_manager_port NODE_MANAGER_PORT]
                 --ray_node_ip_address NODE_IP_ADDRESS
                 [--ray_redis_password REDIS_PASSWORD]
                 [--ray_session_dir SESSION_DIR]
                 [--ray_logs_dir LOGS_DIR]
                 [--runtime-env-hash RUNTIME_ENV_HASH]
                 [--startup_token STARTUP_TOKEN]
```
I'll make an additional PR to make this argument optional so we can have a nice clean merge order

glennmoy (approved):

PR #126, opened 2023-09-18
add/remove local ref on ObjectRef construction/finalize, scaffolding for testing reference counts

kleinschmidt:
Low level JLL: wraps
- `CoreWorker::AddLocalReference`
- `CoreWorker::RemoveLocalReference`
- `CoreWorker::GetAllReferenceCounts`

The last one returns a `std::unordered_map` which requires a custom wrapper for the specific type params (based on the resource request map wrap).

High level Ray API: 
- adds local reference on construction of `ObjectRef` (can be disabled by flag)
- skips adding local reference when ray core worker initializes local ref to be 1 (i.e., `Put` and `SubmitTask`)
- removes local reference in finalizer (cannot be disabled)
- handles `deepcopy` appropriately (routes through constructor in order to increment ref count and install finalizer)

In order to get the high-level support working without too much pain, I decided to chagne teh `ray_jll.ObjectID` (points to C++ managed memory) to the `String`-formatted hex ID, and overloaded the `getproperty` to return a new instance of the `ObjectID` when it's needed (for passing off to Ray C++ code).  I'm not totally happy with this; an alternative would be ot keep the `oid` field and change the handling in the deep copy, but I'm a bit wary about multiple-deallocation/segfaults on finalization...

---
  - https://app.asana.com/0/0/1205510215892649


src/object_ref.jl
```diff
@@ -2,10 +2,33 @@ mutable struct ObjectRef
     oid::ray_jll.ObjectIDAllocated
     owner_address::Union{ray_jll.AddressAllocated,Nothing}
     serialized_object_status::String
+
+    function ObjectRef(oid, owner_address, serialized_object_status;
+                       add_local_ref=true)
+        if add_local_ref
+            worker = ray_jll.GetCoreWorker()
+            ray_jll.AddLocalReference(worker, oid)
+        end
```
omus, line 11: Why not just add the reference after the constructor?

kleinschmidt:
@omus this isn't quite ready for review yet

src/object_ref.jl
```diff
@@ -2,10 +2,33 @@ mutable struct ObjectRef
...
+        objref = new(oid, owner_address, serialized_object_status)
+        return finalizer(objref) do objref
+            # putting finalizer behind `@async` may not be necessary since docs
+            # suggest that you should `ccall` IO functions.  But doing it this
+            # way allows us to do things like debug logging...
+            errormonitor(@async finalize_object_ref(objref))
+            return nothing
+        end
```
omus, line 19: Do you want to conditionally add the `finalizer` based upon `add_local_ref`? Also, since you can add `finalizer`s outside of the constructor it may nicer to separate reference tracking hooks from creating `ObjectRef` instances.

omus (commented):
Took a quick look

src/object_ref.jl
kleinschmidt: (gonna respond to both questions here since they're good ones and related IMO)

I _think_ the only reason to _not_ call `AddLocalReference` during construction is when you get the ID from a core worker operation that _also_ increments the local ref; however, in that case, you _still_ want to _decrement_ the ref when the instance is GCed IIUC.

kleinschmidt: for instance, the object IDs created during task submission are initialized with ref count 1, but the doc string says that the frontend is still responsible for decrementing them when they go out of scope in the application:

```
  /// Add a task that is pending execution.
  ///
  /// The local ref count for all return refs (excluding actor creation tasks)
  /// will be initialized to 1 so that the ref is considered in scope before
  /// returning to the language frontend. The caller is responsible for
  /// decrementing the ref count once the frontend ref has gone out of scope.
  ///
  /// \param[in] caller_address The rpc address of the calling task.
  /// \param[in] spec The spec of the pending task.
  /// \param[in] max_retries Number of times this task may be retried
  /// on failure.
  /// \return ObjectRefs returned by this task.
  std::vector<rpc::ObjectReference> AddPendingTask(const rpc::Address &caller_address,
                                                   const TaskSpecification &spec,
                                                   const std::string &call_site,
                                                   int max_retries = 0);
```

(it's a bit unfortunate that teh `CoreWorker::SubmitTask` docstring does _not_ mention this, you have to dig a bit into [the task manager](https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/core_worker/task_manager.h#L199-L202))

kleinschmidt: Another case: `CoreWorker::Put` calls `AddOwnedReference` and sets the `add_local_ref` arg to `true`:

```
Status CoreWorker::Put(const RayObject &object,
                       const std::vector<ObjectID> &contained_object_ids,
                       ObjectID *object_id) {
  *object_id = ObjectID::FromIndex(worker_context_.GetCurrentInternalTaskId(),
                                   worker_context_.GetNextPutIndex());
  reference_counter_->AddOwnedObject(*object_id,
                                     contained_object_ids,
                                     rpc_address_,
                                     CurrentCallSite(),
                                     object.GetSize(),
                                     /*is_reconstructable=*/false,
                                     /*add_local_ref=*/true,
                                     NodeID::FromBinary(rpc_address_.raylet_id()));
```
https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/core_worker/core_worker.cc#L1110-L1122

```
  /// Add an object that we own. The object may depend on other objects.
  /// Dependencies for each ObjectID must be set at most once. The local
  /// reference count for the ObjectID is set to zero, which assumes that an
  /// ObjectID for it will be created in the language frontend after this call.
  ///
  /// TODO(swang): We could avoid copying the owner_address since
  /// we are the owner, but it is easier to store a copy for now, since the
  /// owner ID will change for workers executing normal tasks and it is
  /// possible to have leftover references after a task has finished.
  ///
  /// \param[in] object_id The ID of the object that we own.
  /// \param[in] contained_ids ObjectIDs that are contained in the object's value.
  /// As long as the object_id is in scope, the inner objects should not be GC'ed.
  /// \param[in] owner_address The address of the object's owner.
  /// \param[in] call_site Description of the call site where the reference was created.
  /// \param[in] object_size Object size if known, otherwise -1;
  /// \param[in] is_reconstructable Whether the object can be reconstructed
  /// through lineage re-execution.
  /// \param[in] add_local_ref Whether to initialize the local ref count to 1.
  /// This is used to ensure that the ref is considered in scope before the
  /// corresponding ObjectRef has been returned to the language frontend.
  /// \param[in] pinned_at_raylet_id The primary location for the object, if it
  /// is already known. This is only used for ray.put calls.
  void AddOwnedObject(const ObjectID &object_id,
                      const std::vector<ObjectID> &contained_ids,
                      const rpc::Address &owner_address,
                      const std::string &call_site,
                      const int64_t object_size,
                      bool is_reconstructable,
                      bool add_local_ref,
                      const absl::optional<NodeID> &pinned_at_raylet_id =
                          absl::optional<NodeID>()) LOCKS_EXCLUDED(mutex_);
```
https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/src/ray/core_worker/reference_count.h#L160-L163

kleinschmidt: in _both_ those cases, we need the finalizer to remove the local ref even though we don't manually add the local ref when that object is created.  I think that, in general, _every_ instance of `ObjectRef` needs to "clean up" after itself like this, but not all of them need to add a local reference on initialization.  So, should we move the add local ref out of hte constructor?  I don't think so, because we don't necessarily control every code path by which an ObjectRef might be created (i.e., copy), so for safety's sake I think its best to default to always adding a local ref on construction.

kleinschmidt:
Something's causing the eval stuff to segfault, will look into it tomorrow.

ray_julia_jll/deps/wrapper.cc
```diff
@@ -475,7 +476,11 @@ JLCXX_MODULE define_julia_module(jlcxx::Module& mod)
     mod.add_type<ObjectID>("ObjectID")
         .method("ObjectIDFromHex", &ObjectID::FromHex)
         .method("ObjectIDFromRandom", &ObjectID::FromRandom)
-        .method("ObjectIDFromNil", &ObjectID::Nil)
+        .method("ObjectIDFromNil", []() {
+            auto id = ObjectID::Nil();
+            ObjectID id_deref = id;
+            return id_deref;
+        })
```
omus, line 483: Should add a test for this

ray_julia_jll/src/wrappers/any.jl
```diff
@@ -159,7 +159,11 @@ NullPtr(::Type{Buffer}) = BufferFromNull()
 
 FromHex(::Type{ObjectID}, str::AbstractString) = ObjectIDFromHex(str)
 FromRandom(::Type{ObjectID}) = ObjectIDFromRandom()
-Nil(::Type{ObjectID}) = ObjectIDFromNil()
+FromNil(::Type{ObjectID}) = ObjectIDFromNil()
```
omus, line 162: I was trying to match the C++ interface with calling this function `Nil`. i.e `ObjectID::FromHex` and `ObjectID::Nil` exist but `ObjectID::FromNil` does not.

ray_julia_jll/src/wrappers/any.jl
```diff
@@ -159,7 +159,11 @@ NullPtr(::Type{Buffer}) = BufferFromNull()
...
+
+Base.show(io::IO, x::ObjectID) = show(io, "ObjectID: " * Hex(x))
+Base.:(==)(a::ObjectID, b::ObjectID) = Hex(a) == Hex(b)
+Base.hash(x::ObjectID, h::UInt) = hash("ObjectID", hash(Hex(x), h))
```
omus, line 166: Doesn't matter much but I think hashing the type is the right way to do this:
```suggestion
Base.hash(x::ObjectID, h::UInt) = hash(ObjectID, hash(Hex(x), h))
```

ray_julia_jll/src/wrappers/any.jl
```diff
@@ -159,7 +159,11 @@ NullPtr(::Type{Buffer}) = BufferFromNull()
 
 FromHex(::Type{ObjectID}, str::AbstractString) = ObjectIDFromHex(str)
 FromRandom(::Type{ObjectID}) = ObjectIDFromRandom()
-Nil(::Type{ObjectID}) = ObjectIDFromNil()
+FromNil(::Type{ObjectID}) = ObjectIDFromNil()
+
+Base.show(io::IO, x::ObjectID) = show(io, "ObjectID: " * Hex(x))
```
omus, line 164: Note that since `repr` falls back to show it's best to output something that can constructed on the REPL:

>  Note that `repr(x)` is usually similar to how the value of x would be entered in Julia

```suggestion
ObjectID(str::AbstractString) = FromHex(ObjectID, str)

Base.show(io::IO, x::ObjectID) = show(io, "ObjectID(\"$(Hex(x))\")")
```

ray_julia_jll/test/reference_counting.jl
```diff
@@ -0,0 +1,26 @@
+@testset "Reference counting" begin
+    using ray_julia_jll: FromRandom, ObjectID, Hex, AddLocalReference, RemoveLocalReference,
+                         GetAllReferenceCounts, _keys, _getindex, GetCoreWorker
+
+    worker = GetCoreWorker()
+    # need to convert to hex string because these are just pointers
+    has_count(oid) = Hex(oid) in Hex.(_keys(GetAllReferenceCounts(worker)))
```
omus, line 7: Is this still true given that `==` was implemented?

ray_julia_jll/test/reference_counting.jl
```diff
@@ -0,0 +1,26 @@
...
+    @test local_count(oid) == 1
+
+    AddLocalReference(worker, oid)
+    @test local_count(oid) == 2
+
+    RemoveLocalReference(worker, oid)
+    @test local_count(oid) == 1
+
+    RemoveLocalReference(worker, oid)
+    @test !has_count(oid)
```
omus, line 25: What happens here with `local_count`?

src/object_ref.jl
```diff
@@ -1,12 +1,60 @@
...
+        objref = new(oid_hex, owner_address, serialized_object_status)
+        if add_local_ref
+            worker = ray_jll.GetCoreWorker()
+            ray_jll.AddLocalReference(worker, objref.oid)
+        end
+        # finalizer(f, object) returns object
+        return finalizer(objref) do objref
+            errormonitor(@async finalize_object_ref(objref))
+            return nothing
+        end
```
omus, line 18: Returning from the `finalizer` makes the code harder to read
```suggestion
       finalizer(objref) do objref
            errormonitor(@async finalize_object_ref(objref))
            return nothing
        end
        return objref
```

src/object_ref.jl
```diff
@@ -1,12 +1,60 @@
...
+        @debug "deep copying x.$(fieldnames[i])"
+        fieldval = getfield(x, fieldnames[i])
+        return Base.deepcopy_internal(fieldval, stackdict)
+    end
+
+    xcp = ObjectRef(fieldcopies...; add_local_ref=true)
+    stackdict[x] = xcp
+
+    return xcp
 end
```
omus, line 53: So the `owner_address` and `serialized_object_status` fields only exist for delaying registration until after deserialization. You can just avoid populating those fields in `deepcopy` as they're only for internal use and we could probably simply this down to just dealing with the `ObjectID`.

src/object_ref.jl
```diff
@@ -1,12 +1,60 @@
 mutable struct ObjectRef
-    oid::ray_jll.ObjectIDAllocated
+    # oid::ray_jll.ObjectIDAllocated
+    oid_hex::String
```
omus, line 3: Can you elaborate on why we should store the hex string instead of the `oid`? maybe we should store both?

test/object_ref.jl
```diff
@@ -27,3 +29,44 @@ end
...
+        oid = obj.oid_hex
+
+        @test local_count(obj) == 1
+
+        obj2 = deepcopy(obj)
+        @test local_count(obj) == 2
+
+        finalize(obj2)
+        # insert yield point here to allow async task that makes the API call to run
+        yield()
```
omus, line 48:
```suggestion
        yield()  # allows async task that makes the API call to run
        
```

ray_julia_jll/deps/wrapper.cc
glennmoy: what's going on here exactly?

ray_julia_jll/test/reference_counting.jl
glennmoy: and what happens if we try to remove it again?

ray_julia_jll/src/wrappers/any.jl
kleinschmidt: Yup, this was intentional: `Nil` returns a _reference_ and it screws up downstream operations with `==`, `hash`, etc.

ray_julia_jll/deps/wrapper.cc
kleinschmidt: `ObjectID::Nil` returns a ref to an `ObjectID` (unlike all the otehr `FromX` functions).  that means taht when comparing two objectIDs we need to handle BOTH values AND refs, and that was even more annoying than I epxected (was getting weird method errors even using `@cxxdereference` because of _slightly_ less generic fallbacks generated by CxxWrap itself).  So I just decided to punt here and return a value even though it's a bit less efficient.

glennmoy: ah I see - thanks!

ray_julia_jll/test/reference_counting.jl
kleinschmidt: the returned dict will not have the corresponding key, so `local_count` gives zero (because it uses `get`)

kleinschmidt: when I was playing around locally, you can remove refs as many times as you want (the count just never goes below zero)

src/object_ref.jl
kleinschmidt: maybe we coudl, but I don't see the downside of honoring the spirit of deepcopy unless there's a very good reason not to.

src/object_ref.jl
kleinschmidt: yup.  was getting some wacky segfaults that went away when I did it this way.  I think probably they were due to double deallocation after copying.  So, we could keep the ObjectID around but we'd need to special case handling of that every time we create an instance.  Even on normal construction, if you construct >1 object ref from a single ObjectID instance, if you don't create a new instance then when you go to finalize the last one living you'll be accessing memory that's already been freed.

Basically it just feels safer to construct an ObjectID on demand when we need it, rather than trying to save a tiny amount of extra allocations by holding onto an instance that's managed by C++.

kleinschmidt:
Getting segfault/deadlock weirdness on #138 (during remove local ref calls in finalizer) even though it seems like the tests all passed, which makes me think the finalizer stuff is a bit fragile unfortunately.  I'm re-running the CI to see if I can get failures to occur here too or if there's something meaningful about the changes made on that branch or if it's general flakiness

kleinschmidt:
The segfaults are definitely flaky/non-deterministic but seem to be happening during the test teardown process, which makes me think there's a race condition between the async objectref finalizers and the core worker cleanup stuff.  The C++ stack trace is different each time which is sus.  I've tried inserting a GC/yield before tearing down the core worker, which seems to have fixed the segfaults on #138 , and cherry picked that to this branch.  The other thing that's probably worth trying at some point is checking whether the core worker is actually initialized in the finalizer, but I'm not sure that would actually solve the problem....

ray_julia_jll/deps/wrapper.cc
omus: Still needs a test

ray_julia_jll/test/reference_counting.jl
omus: Can this be simplified to this now?
```suggestion
    # need to convert to hex string because these are just pointers
    has_count(oid) = oid in _keys(GetAllReferenceCounts(worker))
```

ray_julia_jll/test/reference_counting.jl
omus: Seems like we should test for removing it again:
```suggestion
    @test !has_count(oid)
    @test local_count(oid) == 0
    
    RemoveLocalReference(worker, oid)
    @test !has_count(oid)
    @test local_count(oid) == 0    
```

src/object_ref.jl
```diff
@@ -1,12 +1,60 @@
...
     owner_address::Union{ray_jll.AddressAllocated,Nothing}
```
omus, line 4: Should we keep the string form of this here instead?

src/object_ref.jl
```diff
@@ -1,12 +1,60 @@
 mutable struct ObjectRef
-    oid::ray_jll.ObjectIDAllocated
+    # oid::ray_jll.ObjectIDAllocated
```
omus, line 2:
```suggestion
```

src/object_ref.jl
```diff
@@ -1,12 +1,60 @@
...
     serialized_object_status::String
+
+    function ObjectRef(oid_hex, owner_address, serialized_object_status;
+                       add_local_ref=true)
+        objref = new(oid_hex, owner_address, serialized_object_status)
+        if add_local_ref
+            worker = ray_jll.GetCoreWorker()
+            ray_jll.AddLocalReference(worker, objref.oid)
+        end
+        # finalizer(f, object) returns object
```
omus, line 14:
```suggestion
```

src/object_ref.jl
omus: Still relevant

test/object_ref.jl
```diff
@@ -27,3 +29,44 @@ end
         @test obj_ref1 == obj_ref2
     end
 end
+
+@testset "Local reference counting" begin
+    local_count(o::ObjectRef) = local_count(o.oid_hex)
+    local_count(oid_hex) = first(get(Ray.get_all_reference_counts(), oid_hex, 0))
```
omus, line 35: Seems like this could be implemented in `ray_julia_jll` and we can have a `Ray._local_count` as well

ray_julia_jll/test/reference_counting.jl
omus: Is there any reason to have both `has_count` and `local_count`?

test/utils.jl
```diff
@@ -18,6 +18,9 @@ function setup_core_worker(body)
     try
         body()
     finally
+        # let object ref finalizers run...
+        GC.gc()
+        yield()
```
omus, line 23: Is this required now? What happens if we don't run this?

test/object_ref.jl
```diff
@@ -27,3 +29,44 @@ end
...
+        @test local_count(obj) == 2
+
+        finalize(obj3)
+        yield()
+        @test local_count(obj) == 1
+
+        finalize(obj)
+        yield()
+        @test local_count(oid) == 0
+    end
```
omus, line 61: Should be moved to `object_store.jl` tests due to use of `Ray.put`

test/object_ref.jl
```diff
@@ -27,3 +29,44 @@ end
...
+
+    @testset "Task return object ref" begin
+        obj = Ray.submit_task(getpid, ())
+        oid = obj.oid_hex
+        @test local_count(oid) == 1
+
+        finalize(obj)
+        yield()
+        @test local_count(oid) == 0
+    end
```
omus, line 71: Should be moved to `tasks.jl` tests due to use of `Ray.submit_task`

src/object_ref.jl
kleinschmidt: whoops, thought I got this one

ray_julia_jll/test/reference_counting.jl
kleinschmidt: let's give it a try...

src/object_ref.jl
kleinschmidt: probably but I have those changes on later commits and dont' wanna have to cherry-pick em.  also, that didn't solve the problem `addr_proto.worker_id() != ""` failing

test/utils.jl
kleinschmidt: seg faults!  there's a race condition where the `shutdown_driver` may run before the finalizer tasks run (or even while they are running), causing hte `RemoveLocalReference` segfault when (somewhere in the stack) it refers to memory that's been cleaned up

test/object_ref.jl
kleinschmidt: I moved it to test/utils.jl since it's only really used for testing

ray_julia_jll/deps/wrapper.cc
kleinschmidt: it'll eventually be covered by the tests for teh ownership registration (there weren't any other uses of it or pre-existing tests)

kleinschmidt: ...and none of the other `FromX` have direct tests either

ray_julia_jll/test/reference_counting.jl
kleinschmidt: nope, teh `==` fix is in #138

test/utils.jl
omus: Can you add a comment here about this?

src/object_store.jl
```diff
@@ -78,3 +80,28 @@ function Base.wait(obj_ref::ObjectRef)
...
+    worker = ray_jll.GetCoreWorker()
+    counts_raw = ray_jll.GetAllReferenceCounts(worker)
+
+    # we need to convert this to a dict we can actually work with.  we use the
+    # hex representation of the ID so we can avoid messing with the internal
+    # ObjectID representation...
+    counts = Dict(ray_jll.Hex(k) => Tuple(Int.(ray_jll._getindex(counts_raw, k)))
+                  for k in ray_jll._keys(counts_raw))
+    return counts
+end
```
omus, line 107: Should be moved to `ray_julia_jll`

omus (approved):
Couple of minor things. Also see: https://github.com/beacon-biosignals/Ray.jl/pull/126#discussion_r1334788877

src/object_store.jl
kleinschmidt: why?  we have lots of functions in Ray.jl that wrap API calls like this and provide the high-level julia interface (i.e., `Ray.put`, `Ray.get`, `Ray.get_job_id`, etc.).  as far as I can tell there aren't really clear criteria for what belongs in Ray.jl and what belongs in the jll.

test/utils.jl
kleinschmidt: ah crap I pushed the commit to the wrong branch.  it'll make it in w/ #138 I promise :)

src/object_store.jl
glennmoy: >as far as I can tell there aren't really clear criteria for what belongs in Ray.jl and what belongs in the jll.

we should probably do something about that at some point

omus: In this particular case we mainly use this for:

```julia
local_count(oid_hex) = first(get(Ray.get_all_reference_counts(), oid_hex, 0))
```
Which is something I thought we'd want to reuse in the `ray_julia_jll`.

kleinschmidt: > something I thought we'd want to reuse in the ray_julia_jll

maybe?  I don't know off the top of my head what we'd use it for, but if it does become more convenient to define there we can revisit.

if we do need it, it might be more direct to just directly define that via something like

```julia
local_count(oid_hex) = first(_getindex(GetAllReferenceCounts(GetCoreWorker()), FromHex(ObjectID, oid_hex)))
```

PR #128, opened 2023-09-19
how to install default extra deps from source build

kleinschmidt:
https://stackoverflow.com/questions/30239152/specify-extras-require-with-pip-install-e

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,11 +143,11 @@ Replace the `https://` URL with the approriate release asset URL.
 # https://docs.ray.io/en/releases-2.5.1/ray-contribute/development.html#building-ray-on-linux-macos-full
 python -m pip install --upgrade pip wheel
 cd ray_julia_jll/deps/ray/python
-pip install -e ".[default]" --verbose
+pip install ".[default]" --verbose
 cd -
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,10 +143,12 @@ Replace the `https://` URL with the approriate release asset URL.
 # https://docs.ray.io/en/releases-2.5.1/ray-contribute/development.html#building-ray-on-linux-macos-full
 python -m pip install --upgrade pip wheel
 cd ray_julia_jll/deps/ray/python
-pip install -e . --verbose
+pip install -e ".[default]" --verbose
 cd -
 ```
 
+As with the pre-built wheels, if you don't need to status server/dashboard, you can omit the `[default]` and use `pip install -e . --verbose`.
```
omus, line 150: Do we want to recommend not using `-e` as our default? The `-e` flag "installs a project in editable mode" which most people won't need. I believe I've ran into some issues locally with that flag.
```suggestion
pip install ".[default]" --verbose
cd -
```

As with the pre-built wheels, if you don't need to status server/dashboard, you can omit the `[default]` and use `pip install . --verbose`.
```

omus (approved):

README.md
kleinschmidt: huh, I haven't had any issues with it; I guess the only benefit for using `-e` is if you're editing the actual python files right?

PR #120, opened 2023-09-14
Experimenting with KubeRay support

omus:
TODO:
- [x] Dockerfile: Preserve files created from `copy_to_workspace` rules (see: https://github.com/beacon-biosignals/Ray.jl/pull/120#issuecomment-1722121127)

PR #129, opened 2023-09-19
Pass through metadata into JobConfig

omus:
Improves Julia driver support for Ray job submissions. Fixes this issue on the dashboard where a job submission and the driver are separate entries:

Screenshot 2023-09-18 at 14 30 34
– https://github.com/beacon-biosignals/Ray.jl/pull/120#issuecomment-1724260426

### DIFF
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -97,7 +97,17 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
 
     job_id = ray_jll.GetNextJobID(GLOBAL_STATE_ACCESSOR[])
 
-    job_config = JobConfig(RuntimeEnvInfo(runtime_env))
+    # When submitting a job via `ray job submit` this metadata includes the
+    # "job_submission_id" which lets Ray know that this driver is associated with a
+    # submission ID.
+    metadata = if haskey(ENV, "RAY_JOB_CONFIG_JSON_ENV_VAR")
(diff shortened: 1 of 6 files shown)


omus:
Adding an extra test here to address coverage

src/runtime.jl
```diff
@@ -97,7 +97,17 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
 
     job_id = ray_jll.GetNextJobID(GLOBAL_STATE_ACCESSOR[])
 
-    job_config = JobConfig(RuntimeEnvInfo(runtime_env))
+    # When submitting a job via `ray job submit` this metadata includes the
+    # "job_submission_id" which lets Ray know that this driver is associated with a
+    # submission ID.
+    metadata = if haskey(ENV, "RAY_JOB_CONFIG_JSON_ENV_VAR")
+        json = JSON3.read(ENV["RAY_JOB_CONFIG_JSON_ENV_VAR"])
+        Dict(string(k) => v for (k, v) in json.metadata)
```
omus, line 105: Required due to:
```
julia> Dict{String,String}(:key => "value")
ERROR: MethodError: Cannot `convert` an object of type Symbol to an object of type String
```
I'd prefer to leave `JobConfig.metadata::Dict{String,String}` as this matches the C++ better

kleinschmidt (approved):

PR #130, opened 2023-09-19
Create Ray.jl `Dockerfile`

omus:
Creates an Ray.jl compatible alternative to the image `rayproject/ray:2.5.1-py310`. Additionally, this image features:

- Julia 1.9.3 pre-installed
- Ray.jl and ray_julia_jll.jl pre-installed based upon version in worktree
- Ray CLI from our https://github.com/beacon-biosignals/ray configurable via a build argument
- Docker caching for faster subsequent builds of ray_julia_jll and Ray CLI

There are additional tweaks that can be done here but the `Dockerfile` in its current form is functional and decently fast on subsequent builds. The image is currently x86_64 only at the moment as the `rayproject/ray` image requires the use of a separate tag for aarch64.

Uses the commit from https://github.com/beacon-biosignals/ray/pull/10. Ideally we'd merge that PR first and update the hardcoded Dockerfile commit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,196 @@
+# syntax=docker/dockerfile:1
+
+# TODO: Cleanup uid/gid/user work arounds
+
+ARG JULIA_VERSION=1.9.3
+ARG RAY_VERSION=2.5.1
+
+FROM julia:${JULIA_VERSION}-bullseye as julia-base
+
+# Based upon `/etc/debian_version` the `ray:2.5.1` image is based on Debian Bullseye.
+# No automatic multi-architecture support at the moment. Must specify `-aarch64` suffix
+# otherwise the default is x86_64 (https://github.com/ray-project/ray/tree/master/docker/ray#tags)
+FROM rayproject/ray:${RAY_VERSION}-py310 as ray-base
+
+# Install Julia
+COPY --link --from=julia-base /usr/local/julia /usr/local/julia
+ENV JULIA_PATH /usr/local/julia
+ENV PATH $JULIA_PATH/bin:$PATH
+
+# Validate Julia executable is compatible with the container architecture
+RUN if ! julia --history-file=no -e 'exit(0)'; then \
+        uname -m && \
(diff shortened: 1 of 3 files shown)


ray_julia_jll/deps/build.jl
```diff
@@ -20,7 +20,8 @@ open(joinpath(build_dir, "WORKSPACE.bazel"), "w+") do io
     Mustache.render(io, template, dict)
 end
 
-isdir(ray_dir) || cd(dirname(ray_dir)) do
+# Clone "ray" repo when the directory is missing or empty
+isdir(ray_dir) && !isempty(readdir(ray_dir)) || cd(dirname(ray_dir)) do
```
omus, line 24: The `Dockerfile` doesn't currently utilize this new feature but this was handy when I was precreating the `ray_julia_jll/deps/ray` directory with a symlink.

omus:
We probably want to automatically build this Dockerfile as part of CI. I can definitely do so but may want to punt on it for the time being. @kleinschmidt let me know if that's okay with you.

Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+    fi && \
+    #
+    # Build the dashboard
+    cd ${JLL_JULIA_PROJECT}/deps/ray/python/ray/dashboard/client && \
+    npm ci && \
+    npm run build && \
+    #
+    # Build Ray for Python
+    cd ${JLL_JULIA_PROJECT}/deps/ray/python && \
+    pip install --verbose . && \
```
kleinschmidt, line 154: I needed to install the additioanl "default" dependencies in order to be able to use things like `ray list objects`...is there a reason not to do that here?

```suggestion
    pip install --verbose ".[default]" && \
```

kleinschmidt (approved):

Dockerfile
omus: It doesn't appear to be required to run jobs on the cluster. Possibly it would be useful if you are using the Ray CLI from within the cluster but this will take additional time to build and I doubt we'll use these features.

PR #138, opened 2023-09-20
include nested IDs in put/task args/returns

kleinschmidt:
WIP, builds on #126 

Adds a centralized `serialize_to_ray_object` function which creates a `ray::core::RayObject` holding data, metadata, and nested object refs (`rpc::ObjectReference`).  Then extracts the `ObjectID`s from that when doing `put`, serializing task args, and generating the return values.

Haven't tackled de-serialization yet (including testing for the return values ref counting properly), but will leave that for a follow-up, as it requires tackling `RegisterOwnershipAndResolveFuture` I think.

---
  - https://app.asana.com/0/0/1205510215892650

### DIFF
--- a/ray_julia_jll/src/wrappers/any.jl
+++ b/ray_julia_jll/src/wrappers/any.jl
@@ -2,6 +2,7 @@ using CxxWrap
 using CxxWrap.StdLib: StdVector, SharedPtr
 using libcxxwrap_julia_jll
 
+using InteractiveUtils: subtypes
 using Serialization
 
 JLLWrappers.@generate_wrapper_header("ray_julia")
@@ -167,8 +168,26 @@ FromNil(::Type{ObjectID}) = ObjectIDFromNil()
 
 ObjectID(str::AbstractString) = FromHex(ObjectID, str)
 
-Base.show(io::IO, x::ObjectID) = show(io, "ObjectID(\"$(Hex(x))\")")
-Base.:(==)(a::ObjectID, b::ObjectID) = Hex(a) == Hex(b)
+Base.show(io::IO, x::ObjectID) = write(io, "ObjectID(\"", Hex(x), "\")")
+
+# cannot believe I'm doing this...
+# 
+# Because ObjectID is a CxxWrap-defined type, it has two subtypes:
+# `ObjectIDAllocated` and `ObjectIDDereferenced`.  The first is returned when we
+# construct directly or return by value, the second when you pull a ref out of
+# say `std::vector<ObjectID>`.
+#
+# ObjectID is abstract, so the normal method definition:
+# 
+# Base.:(==)(a::ObjectID, b::ObjectID) = Hex(a) == Hex(b)
+#
+# is shadowed by more specific fallbacks defined by CxxWrap.
+for Ta in subtypes(ObjectID)
+    for Tb in subtypes(ObjectID)
+        @eval Base.:(==)(a::$Ta, b::$Tb) = Hex(a) == Hex(b)
+    end
+end
+
 Base.hash(x::ObjectID, h::UInt) = hash(ObjectID, hash(Hex(x), h))
 
 #####
--- a/ray_julia_jll/src/wrappers/any.jl
+++ b/ray_julia_jll/src/wrappers/any.jl
@@ -2,7 +2,6 @@ using CxxWrap
 using CxxWrap.StdLib: StdVector, SharedPtr
 using libcxxwrap_julia_jll
 
-using InteractiveUtils: subtypes
 using Serialization
 
 JLLWrappers.@generate_wrapper_header("ray_julia")
@@ -182,10 +181,9 @@ Base.show(io::IO, x::ObjectID) = write(io, "ObjectID(\"", Hex(x), "\")")
 # Base.:(==)(a::ObjectID, b::ObjectID) = Hex(a) == Hex(b)
 #
 # is shadowed by more specific fallbacks defined by CxxWrap.
-for Ta in subtypes(ObjectID)
-    for Tb in subtypes(ObjectID)
-        @eval Base.:(==)(a::$Ta, b::$Tb) = Hex(a) == Hex(b)
-    end
+const ObjectIDTypes = (ObjectIDAllocated, ObjectIDDereferenced)
+for Ta in ObjectIDTypes, Tb in ObjectIDTypes
+    @eval Base.:(==)(a::$Ta, b::$Tb) = Hex(a) == Hex(b)
 end
 
 Base.hash(x::ObjectID, h::UInt) = hash(ObjectID, hash(Hex(x), h))
--- a/ray_julia_jll/test/objectid.jl
+++ b/ray_julia_jll/test/objectid.jl
@@ -1,12 +1,7 @@
 @testset "ObjectID" begin
     using ray_julia_jll: ObjectID, FromRandom, FromHex, Hex
-    # need to get some ObjectIDDereferenced
-    objs = CxxWrap.StdVector{ObjectID}()
     obj = FromRandom(ObjectID)
-    push!(objs, obj)
-    push!(objs, obj)
-
-    obj_deref = objs[1]
+    obj_deref = CxxRef(obj)[]
     @test obj_deref isa ray_julia_jll.ObjectIDDereferenced
     @test obj isa ray_julia_jll.ObjectIDAllocated
 
@@ -17,7 +12,7 @@
     @test obj2 !== obj
     @test obj2 == obj
 
-    obj_deref2 = objs[2]
+    obj_deref2 = CxxRef(obj2)[]
     @test obj_deref2 !== obj_deref
     # confirm C++ pointers are different too
     @test obj_deref2.cpp_object != obj_deref.cpp_object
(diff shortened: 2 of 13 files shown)


src/ray_serializer.jl
```diff
@@ -52,4 +52,20 @@ function serialize_to_bytes(x)
...
 end
 
+function serialize_to_ray_object(x)
+    bytes = Vector{UInt8}()
+    s = RaySerializer(bytes)
+    writeheader(s)
+    serialize(s, x)
+
+    buffer = ray_jll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
+    metadata = ray_jll.NullPtr(ray_jll.Buffer)
```
glennmoy, line 62: do we know yet what metadata is supposed to hold?

kleinschmidt: we have some hunches yes :)  centralizing the "serialize this thing to a RayObject" in a single function means there's a centralized place where we can handle the metadata too, but we're not yet at a point where we can tackle that.

ray_julia_jll/src/wrappers/any.jl
```diff
@@ -167,8 +168,26 @@ FromNil(::Type{ObjectID}) = ObjectIDFromNil()
...
+# ObjectID is abstract, so the normal method definition:
+# 
+# Base.:(==)(a::ObjectID, b::ObjectID) = Hex(a) == Hex(b)
+#
+# is shadowed by more specific fallbacks defined by CxxWrap.
+for Ta in subtypes(ObjectID)
+    for Tb in subtypes(ObjectID)
+        @eval Base.:(==)(a::$Ta, b::$Tb) = Hex(a) == Hex(b)
+    end
+end
```
omus, line 189: Aren't we invalidating some CxxWrap code by doing this? Seems like it's better just to avoid defining this and keeping the nice `==` in `ObjectRef`.

Also, does this work:
```julia
Base.:(==)(a::T1, b::T2) where T1 <: ObjectID, T2 <: ObjectID = Hex(a) == Hex(b)
```

ray_julia_jll/test/objectid.jl
```diff
@@ -0,0 +1,29 @@
+@testset "ObjectID" begin
+    using ray_julia_jll: ObjectID, FromRandom, FromHex, Hex
+    # need to get some ObjectIDDereferenced
+    objs = CxxWrap.StdVector{ObjectID}()
+    obj = FromRandom(ObjectID)
+    push!(objs, obj)
+    push!(objs, obj)
+
+    obj_deref = objs[1]
```
omus, line 9: A better way:
```julia
julia> using ray_julia_jll, CxxWrap

julia> x = ray_julia_jll.FromRandom(ray_julia_jll.ObjectID)
ray_julia_jll.ObjectIDAllocated(Ptr{Nothing} @0x000060000311eeb0)

julia> CxxRef(x)[]
ray_julia_jll.ObjectIDDereferenced(Ptr{Nothing} @0x000060000311eeb0)
```

```suggestion
    obj = FromRandom(ObjectID)
    obj_deref = CxxRef(obj)[]
```

ray_julia_jll/test/objectid.jl
```diff
@@ -0,0 +1,29 @@
...
+    @test obj isa ray_julia_jll.ObjectIDAllocated
+
+    @test obj_deref == obj
+    @test obj == obj_deref
+
+    obj2 = FromHex(ObjectID, Hex(obj))
+    @test obj2 !== obj
+    @test obj2 == obj
+
+    obj_deref2 = objs[2]
```
omus, line 20:
```suggestion
    obj_deref2 = CxxRef(obj2)[]
```

src/object_store.jl
```diff
@@ -5,11 +5,14 @@ Store `data` in the object store. Returns an object reference which can used to
 the `data` with [`Ray.get`](@ref).
 """
 function put(data)
-    bytes = serialize_to_bytes(data)
-    buffer = ray_jll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
-    ray_obj = ray_jll.RayObject(buffer)
+    # bytes = serialize_to_bytes(data)
+    # buffer = ray_jll.LocalMemoryBuffer(bytes, sizeof(bytes), true)
+    # ray_obj = ray_jll.RayObject(buffer)
```
omus, line 10:
```suggestion
```

src/runtime.jl
```diff
@@ -238,17 +238,20 @@ function serialize_args(args)
...
+        # serialized_arg_size = sizeof(serialized_arg)
+
+        # buffer = ray_jll.LocalMemoryBuffer(serialized_arg, serialized_arg_size, true)
+        # metadata = ray_jll.NullPtr(ray_jll.Buffer)
+        # inlined_ids = StdVector(collect(serializer.object_ids))::StdVector{ray_jll.ObjectID}
+        # inlined_refs = ray_jll.GetObjectRefs(worker, inlined_ids)
+        # ray_obj = ray_jll.RayObject(buffer, metadata, inlined_refs, false)
+
+        ray_obj = serialize_to_ray_object(arg)
+        serialized_arg_size = ray_jll.GetSize(ray_obj[])
```
omus, line 254: Nicely done!

src/runtime.jl
```diff
@@ -238,17 +238,20 @@ function serialize_args(args)
...
+        # writeheader(serializer)
+        # serialize(serializer, arg)
+        # serialized_arg_size = sizeof(serialized_arg)
+
+        # buffer = ray_jll.LocalMemoryBuffer(serialized_arg, serialized_arg_size, true)
+        # metadata = ray_jll.NullPtr(ray_jll.Buffer)
+        # inlined_ids = StdVector(collect(serializer.object_ids))::StdVector{ray_jll.ObjectID}
+        # inlined_refs = ray_jll.GetObjectRefs(worker, inlined_ids)
+        # ray_obj = ray_jll.RayObject(buffer, metadata, inlined_refs, false)
+
```
omus, line 252:
```suggestion
```

test/object_ref.jl
```diff
@@ -69,4 +72,17 @@ end
...
     end
+
+    @testset "Task argument object ref" begin
+        obj = Ray.put(1)
+        # XXX: The "task dependency" ref count goes to 0 when the task is
+        # complete; Ray tests use an actor task which can be triggered to
+        # complete by sending a "signal" (via another task call) to avoid race
+        # conditions.  Here we're just hoping that the following test runs
+        # within ~3s.
+        ret_obj = Ray.submit_task(x -> (Sleep(3); Ray.get(x)), (obj,))
```
omus, line 83: Why isn't this failing?
```suggestion
        ret_obj = Ray.submit_task(x -> (sleep(3); Ray.get(x)), (obj,))
```

test/object_ref.jl
```diff
@@ -69,4 +72,17 @@ end
...
+        @test !isready(ret_obj) && task_dep_count(obj) == 1
+        wait(ret_obj)
+        @test task_dep_count(obj) == 0
+    end
```
omus, line 87: I've been putting similar tests in `task.jl`. Basically my rule is if it calls `Ray.submit_task` it goes in the tasks.jl tests

test/ray_serializer.jl
```diff
@@ -77,3 +77,25 @@ end
...
+
+        stuff2 = [obj1, (obj2, obj1, "blah"), 1]
+        ray_obj2 = Ray.serialize_to_ray_object(stuff2)
+        nested_obj_ids2 = ray_jll.GetNestedRefIds(ray_obj2[])
+        @test length(nested_obj_ids2) == 2
+        @test issetequal(nested_obj_ids2, nested_obj_ids)
+
+        @test Ray.get(ray_obj2) == stuff2
+    end
+end
```
omus, line 101: Seems like more of an `object_store.jl` test. Is there a way to test this without calling `get`/`put` here?

ray_julia_jll/src/wrappers/any.jl
kleinschmidt: that's equivalent to `Base.:(==)(a::ObjectID, b::ObjectID)` right?

I'm not sure if we're invalidating CxxWrap-generated code, any more than we would be defining the original method.

test/object_ref.jl
kleinschmidt: 😬

test/ray_serializer.jl
kleinschmidt: in this particular case I think we could get away with random ones.  but once we start trying to register ownership, I think things will blow up (need to be a "real" object ref, not a random one for that)

ray_julia_jll/src/wrappers/any.jl
omus, line 189: Probably should do:
```suggestion
for Ta in subtypes(ObjectID), Tb in subtypes(ObjectID)
    @eval Base.:(==)(a::$Ta, b::$Tb) = Hex(a) == Hex(b)
end
```

If we want to drop the extra dependency I'd be good with doing:
```suggestion
for Ta in (ObjectIDAllocated, ObjectIDDereferenced), Tb in (ObjectIDAllocated, ObjectIDDereferenced)
    @eval Base.:(==)(a::$Ta, b::$Tb) = Hex(a) == Hex(b)
end
```
I feel like CxxWrap is clear about the naming of these subtypes so we don't need to do it dynamically. Also, if an abstract subtype is added this may no longer be specific enough again.

omus (approved):

PR #139, opened 2023-09-21
Refactor `Dockerfile` to be useful as a base image for images

omus:
Did a bunch of refactoring as part of this. The main change here was installing the Ray.jl package in `~/.julia/dev/Ray` to make it easier to access from other images. Additionally, this Dockerfile doesn't rely on using `JULIA_PROJECT` and instead adds Ray.jl to the default Julia environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,8 @@ FROM rayproject/ray:${RAY_VERSION}-py310 as ray-base
 
 # Install Julia
 COPY --link --from=julia-base /usr/local/julia /usr/local/julia
-ENV JULIA_PATH /usr/local/julia
-ENV PATH $JULIA_PATH/bin:$PATH
+ENV JULIA_PATH=/usr/local/julia
+ENV PATH=$JULIA_PATH/bin:$PATH
 
 # Validate Julia executable is compatible with the container architecture
 RUN if ! julia --history-file=no -e 'exit(0)'; then \
@@ -39,7 +39,7 @@ ENV JULIA_CPU_TARGET="generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,bas
 FROM ray-base as deps
 
 # Disable automatic package precompilation. We'll control when packages are precompiled.
-ENV JULIA_PKG_PRECOMPILE_AUTO "0"
+ENV JULIA_PKG_PRECOMPILE_AUTO="0"
 
 # Use the Git CLI when we are not using the Julia's PkgServer as otherwise Docker image
 # cross compilation can cause LibGit2 to run out of memory while cloning the General registry.
@@ -58,7 +58,7 @@ RUN --mount=type=cache,sharing=locked,target=/tmp/julia-cache,uid=1000,gid=100 \
     julia -e 'using Pkg; Pkg.Registry.add("General")'
 
 # Instantiate the Julia project environment
-ENV JULIA_PROJECT /Ray.jl
+ENV JULIA_PROJECT=/Ray.jl
 COPY --chown=ray *Project.toml *Manifest.toml ${JULIA_PROJECT}/
 
 # Generate a fake ray_julia_jll package just for instantiation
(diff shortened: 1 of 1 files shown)


omus:
As I used this code in the https://github.com/beacon-biosignals/data-map/pull/1371#issuecomment-1732011163 benchmarks I make additional changes as a follow up

omus:
As I used this code in some private benchmarks I'd like to merge these changes mostly as is and make additional changes as a follow up

Dockerfile
```diff
@@ -49,31 +54,33 @@ ENV JULIA_PKG_USE_CLI_GIT="true"
...
-RUN --mount=type=cache,sharing=locked,target=/tmp/julia-cache,uid=1000,gid=100 \
-    julia -e 'using Pkg; Pkg.Registry.update(); Pkg.instantiate(); Pkg.build(); Pkg.precompile(strict=true, timing=true)'
+ARG RAY_JL_PROJECT=${HOME}/.julia/dev/Ray
+COPY --chown=ray *Project.toml *Manifest.toml /tmp/Ray.jl/
+RUN --mount=type=cache,sharing=locked,target=${JULIA_DEPOT_CACHE},uid=${UID},gid=${GID} \
+    # Move project content into temporary depot
+    rm -rf ${RAY_JL_PROJECT} && \
+    mv /tmp/Ray.jl ${RAY_JL_PROJECT} && \
+    # Generate a fake ray_julia_jll package just for instantiation
+    mkdir -p ${RAY_JL_PROJECT}/ray_julia_jll/src && touch ${RAY_JL_PROJECT}/ray_julia_jll/src/ray_julia_jll.jl && \
```
kleinschmidt, line 73: thsi will need to change w/ #159 etc. right?

Dockerfile
```diff
@@ -49,31 +54,33 @@ ENV JULIA_PKG_USE_CLI_GIT="true"
...
+    # Note: The `timing` flag requires Julia 1.9
+    julia --project=${RAY_JL_PROJECT} -e 'using Pkg; Pkg.Registry.update(); Pkg.instantiate(); Pkg.build(); Pkg.precompile(strict=true, timing=true)'
```
kleinschmidt, line 75: we no longer are building from source by default right?  Pkg.build shoudl be a no-op then, or do we intend to still actually build the JLLs?

kleinschmidt: ah right at this stage, we're not doing any kind of real build because we've created a fake (empty) jll subdir.  however, when we swithc to making it a submodule, will this still work as intended?

Dockerfile
```diff
@@ -104,97 +112,102 @@ RUN sudo ln -s ../lib/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm
...
 # Copy over artifacts generated during the previous stages
 COPY --chown=ray --link --from=deps ${HOME}/.julia ${HOME}/.julia
 
 # Setup ray_julia_jll
-ARG JLL_ROOT=/ray_julia_jll
-COPY --chown=ray ray_julia_jll ${JLL_ROOT}
-RUN --mount=type=cache,sharing=locked,target=/mnt/bazel-cache,uid=1000,gid=100 \
+ARG RAY_JLL_REPO=${HOME}/.julia/dev/ray_julia_jll
+COPY --chown=ray ray_julia_jll ${RAY_JLL_REPO}
+RUN --mount=type=cache,sharing=locked,target=${BAZEL_CACHE},uid=${UID},gid=${GID} \
```
kleinschmidt, line 177: this is gonna change with #158 too right?

kleinschmidt (approved):

Dockerfile
kleinschmidt: er, I mean #158

omus: Yup, hence why I'm wanting to get this merged in

Dockerfile
omus: So `Pkg.build` is a no-op currently but it's best to include it in case a package depended on by Ray.jl requires it. 

> ah right at this stage, we're not doing any kind of real build because we've created a fake (empty) jll subdir. however, when we swithc to making it a submodule, will this still work as intended?

I'll need to do a couple of changes in the Dockerfile to accommodate the submodule refactor. I'll just be able to delete the creation of the fake JLL in this case

Dockerfile
omus: Yup, will need to update all the paths here

PR #145, opened 2023-09-25
Apply YAS formatting

omus:
Should be applying the YAS style guide before our first real release.

### DIFF
--- a/src/Ray.jl
+++ b/src/Ray.jl
@@ -14,7 +14,7 @@ using Logging
 using LoggingExtras
 using Pkg
 using Serialization: Serialization, AbstractSerializer, Serializer, deserialize,
-    reset_state, serialize, serialize_type, ser_version, writeheader
+                     reset_state, serialize, serialize_type, ser_version, writeheader
 using Sockets: IPAddr, getipaddr
 
 export start_worker, submit_task, @ray_import, ObjectRef
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,1 @@
+style = "yas"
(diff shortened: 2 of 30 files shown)


src/Ray.jl
```diff
@@ -14,7 +14,7 @@ using Logging
 using LoggingExtras
 using Pkg
 using Serialization: Serialization, AbstractSerializer, Serializer, deserialize,
-    reset_state, serialize, serialize_type, ser_version, writeheader
+                     reset_state, serialize, serialize_type, ser_version, writeheader
 using Sockets: IPAddr, getipaddr
 
 import ray_julia_jll as ray_jll
```
ararslan, line 20:
```suggestion
using ray_julia_jll: ray_julia_jll as ray_jll
```
Looks like JuliaFormatter doesn't recognize this case

omus: TIL

omus:
> I assume the formatter was applied here so I'll pre-approve to not block the release

Yes it was. I've also introduced a format check workflow to enforce it

kleinschmidt:
> I've also introduced a format check workflow to enforce it

The auto-formatter action isn't very useful, it just tells you if it failed and which file, but doesn't suggest how to fix it.  It's also hard to run locally (you need to setup your own environmetn with formatter and run it in the same way that the action does.  Better IMO would be to use a separate `/format` project that has a `run.jl` script so you can run it locally in exactly the same way.  The action also doesn't specify what format to use.  I'll make a follow up PR to implement these suggestions!

kleinschmidt:
Ah I see now, it's specified in the JuliaFormatter.toml file

PR #146, opened 2023-09-25
Use `cache-compiled` feature from `julia-actions/cache`

omus:
Uses a feature from this GitHub action to allow caching of Julia `.ji` files. Upstream issue: https://github.com/julia-actions/cache/issues/11


kleinschmidt (approved):
including the os/arch in the cache name handles my one concern (if we start doing multi-arch builds)

PR #150, opened 2023-09-26
Use a separate project for `ray_julia_jll/deps`

omus:
Doing this avoids having Ray.jl users take on additional dependencies only needed for building the package from source. With this change we're also no longer using the pre-built binaries in CI. I think we should probably have tests for both building from source or from pre-built binaries but that should be a separate job/workflow. I'll do that as a follow up PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -112,11 +112,6 @@ jobs:
           using Pkg
           Pkg.instantiate()
           include("${{ env.ray_julia_jll_dir }}/deps/build_jll.jl")
-
-          cat >> ~/.julia/artifacts/Overrides.toml <<-EOF
-          [c348cde4-7f22-4730-83d8-6959fb7a17ba]
-          ray_julia = "$(pwd)/ray_julia_jll/deps/bazel-bin"
-          EOF
       - name: "Monorepo setup"
         shell: julia --color=yes --project {0}
         run: |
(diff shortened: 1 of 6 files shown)


ray_julia_jll/deps/build_jll.jl
```diff
@@ -32,3 +32,13 @@ end
 if !isfile(joinpath(artifact_dir, library_name))
     error("Failed to build library: $(joinpath(artifact_dir, library_name))")
 end
+
+# Add entry to depot Overrides.toml
+if "--no-override" in ARGS
```
glennmoy, line 37:
```suggestion
if !("--no-override" in ARGS)
```

.github/workflows/CI.yml
```diff
@@ -106,11 +107,16 @@ jobs:
           mkdir -p "$RAY_GEN_CACHE_DIR"
           cp -rfp . "$RAY_GEN_CACHE_DIR"
         working-directory: ${{ env.ray_dir }}
+      - name: Build ray_julia library
+        shell: julia --color=yes --project {0}
+        run: |
+          using Pkg
+          Pkg.instantiate()
+          include(joinpath(pwd(), "build_jll.jl"))
```
omus, line 115: Using `include("build_jll.jl")` doesn't work for some reason. We can look into that later as we need to get the CI building from source again so we can get #140 in.

glennmoy (approved):

PR #154, opened 2023-09-27
Install Ray from prebuilt wheels in CI

glennmoy:


omus (commented):
I'm not sure we want to do this for the from source CI workflow. The rational is that if you make a Ray.jl PR that requires changes from https://github.com/beacon-biosignals/ray then the PR cannot be tested until the wheel is updated.

What I could get behind is having our Ray.jl code base use a fixed version of https://github.com/beacon-biosignals/ray that is tracked in the repo history. Having this would allow us to make reproducible builds from source for older Ray.jl commits.

glennmoy:
> I'm not sure we want to do this for the from source CI workflow. The rational is that if you make a Ray.jl PR that requires changes from https://github.com/beacon-biosignals/ray then the PR cannot be tested until the wheel is updated.

Makes sense, I guess it'd be annoying if we had to update the wheels several times before the CI passed...

We already use a fixed wheel in the Artifacts workflow anyway, so it's tested for when a user is involved.

> What I could get behind is having our Ray.jl code base use a fixed version of https://github.com/beacon-biosignals/ray that is tracked in the repo history. Having this would allow us to make reproducible builds from source for older Ray.jl commits.

Yeah I think we should do that. I guess we should set this in some `requirements.txt`?

glennmoy:
Closing in favour of #164

PR #155, opened 2023-09-27
Add CI workflow to test Dockerfile build

omus:
Follow up to: https://github.com/beacon-biosignals/Ray.jl/pull/130#issuecomment-1726025172


omus:
I'll get back to this PR after the submodule refactor: #158

.github/workflows/Dockerfile.yml
```diff
@@ -0,0 +1,41 @@
...
+      # Reference the HEAD commit which triggered this workflow. By default PRs use a merge commit
+      SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ env.SHA }}  # Always checkout HEAD commit
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Build
+        uses: docker/build-push-action@v4
```
glennmoy, line 32: any reason not to use v5?

.github/workflows/Dockerfile.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          # Use Path context to ensure that the `.git` directory is included in the image.
+          # https://github.com/docker/build-push-action#git-context
+          context: .
+          push: false
+          file: Dockerfile
+          # TODO: May want to switch to another cache type as GHA have a 10 GB limit
```
glennmoy, line 39: what's the current image size out of interest? would there be any utility in printing that to CI so we can see if we're approaching the limit? (I can see the individual layher sizes in the CI output but not a total image size from what I can see)

glennmoy (approved):

.github/workflows/Dockerfile.yml
omus: Not as far as I know

.github/workflows/Dockerfile.yml
omus: The image is currently 5.93GB (from my local build). I'll try to add that info to this workflow. I'll call out that we are allowed to go over the limit but GHA will eventually clean up older artifacts

.github/workflows/Dockerfile.yml
```diff
@@ -43,4 +43,4 @@ jobs:
       - name: Report Docker image size
         run: |
           size=$(docker images ray-jl --format "{{.Size}}")
-          echo "Docker image size: $size" | tee -a "$GITHUB_STEP_SUMMARY"
+          echo ":whale: Docker image size: $size" | tee -a "$GITHUB_STEP_SUMMARY"
```
glennmoy, line 47: nice touch 😂

PR #158, opened 2023-09-27
Refactor `ray_julia_jll` into a submodule

omus:
Converts `ray_julia_jll` from a package to a submodule in Ray.jl. See #124 for additional context.

PR #160 should be merged first.

PR #159, opened 2023-09-27
Combine `ray_julia_jll` deps/gen directories

omus:
While working on #158 I decided this was something we should also tackle. The rational behind this is that the `deps` directory in Julia is usually reserved for runtime dependencies of the package that are automatically built during package installation. As we're moving towards using artifacts only developers of Ray.jl will be required to build the "ray_julia" shared library and utilize `~/.julia/artifacts/Overrides.toml`. Due to this it seems best to combine the current artifact building directory `gen` and `deps` together into a new directory: `build`.

I also considered using the directory name `artifacts` but thought it could confuse users into thinking that the artifacts were embedded into the package.


omus:
I was going to include this change in #158 but decided against it as both of these PRs include a non-trivial amount of changes.

ray_julia_jll/build/build.jl
kleinschmidt, line 1: I wonder if we want to keep this one as `build_jll` given that there are other `build_*` in here?

ray_julia_jll/build/build_tarballs.jl
```diff
@@ -19,13 +19,13 @@ if abspath(PROGRAM_FILE) == @__FILE__
     isdir(TARBALL_DIR) || mkdir(TARBALL_DIR)
 
     @info "Building ray_julia_jll..."
-    run(`$(Base.julia_cmd()) --project=$JLL_PATH $JLL_PATH/deps/build_jll.jl `)
+    include("build.jl")
```
kleinschmidt, line 22: are we sure that we don't need to fork a new julia process to run this build step?  why were we doing that in the first place?

Dockerfile
```diff
@@ -178,22 +178,22 @@ RUN --mount=type=cache,sharing=locked,target=${BAZEL_CACHE},uid=${UID},gid=${GID
...
     #
     # Build using the final Ray.jl destination
-    ln -s ${RAY_REPO} ${RAY_JLL_REPO}/deps/ray && \
+    ln -s ${RAY_REPO} ${RAY_JLL_REPO}/build/ray && \
     rm -rf ${RAY_JLL_PROJECT} && \
     ln -s ${RAY_JLL_REPO} ${RAY_JLL_PROJECT} && \
     #
     # Build ray_julia_jll
-    julia --project=${RAY_JLL_PROJECT} -e 'using Pkg; Pkg.build(verbose=true); Pkg.precompile(strict=true)' && \
+    julia --project=${RAY_JLL_PROJECT} -e 'using Pkg; Pkg.build(); Pkg.precompile(strict=true)' && \
```
kleinschmidt, line 186: can this still be done w/ `Pkg.build` or do we need to actually `include` the build.jl file?

ray_julia_jll/.gitignore
```diff
@@ -1,9 +1,7 @@
...
-deps/build.log
-deps/WORKSPACE.bazel
-deps/ray
-deps/bazel-*
-deps/override
 /.dir-locals.el
-gen/tarballs
+build/WORKSPACE.bazel
+build/ray
+buid/bazel-*
```
kleinschmidt, line 6:
```suggestion
build/bazel-*
```

ray_julia_jll/build/README.md
kleinschmidt, line 1: this needs to be updated as well with the changes to the filenames (make.jl -> build_tarballs.jl) for instance

kleinschmidt (approved):
a few Qs and small things but otherwise looks good

ray_julia_jll/build/build.jl
omus: I think `build_library.jl` would be the name to use if we did this

ray_julia_jll/build/build_tarballs.jl
omus: The reason we previously did this was due to the `build_jll.jl` and the the `make.jl` having separate projects with different dependencies.

Dockerfile
omus: The `Dockerfile` needs to include the `build.jl` now and should have included the `build_jll.jl` when that was added. Having #155 would help catch these things in the future. I'll be making another PR which fixes the Dockerfile

omus: Opened #160 for this. As the Dockerfile is currently broken on `main` I think I'll get this merged so we can keep things moving.

PR #160, opened 2023-09-27
Add `BUILD_PROJECT` argument to Dockerfile

omus:
We've been heavily modifying our build setup and our Dockerfile has fallen out of date. I've added the `BUILD_PROJECT` Docker build arg to help with this (easier to keep things up to date).

Additionally, we're now properly calling `build_jll.jl`


omus:
Updated this PR to work with #159. I've manually confirmed that I can build the image.

glennmoy (approved):

PR #164, opened 2023-09-28
Use fixed version of ray-project

omus:
In response to: https://github.com/beacon-biosignals/Ray.jl/pull/154#pullrequestreview-1649351425

By using a specific commit of `ray` in our `build_library.jl` script we can ensure that from source builds are full reproducible from any commit (from now on). We could just use the `WORKSPACE.bazel` to also do this but this allows us the flexibility to develop against a branch via `--no-checkout`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -122,12 +122,12 @@ ARG BUILD_PROJECT=${RAY_JL_PROJECT}/build
 ARG RAY_REPO=${HOME}/ray
 ARG RAY_REPO_CACHE=/mnt/ray-cache
 ARG RAY_CACHE_CLEAR=false
-COPY --chown=${UID} build/ray-commit /tmp/ray-commit
+COPY --chown=${UID} build/ray_commit /tmp/ray_commit
 RUN --mount=type=cache,sharing=locked,target=${BAZEL_CACHE},uid=${UID},gid=${GID} \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -120,14 +120,14 @@ ARG BUILD_PROJECT=${RAY_JL_PROJECT}/build
 # Install custom Ray CLI which supports the Julia language.
 # https://docs.ray.io/en/releases-2.5.1/ray-contribute/development.html#building-ray-on-linux-macos-full
 ARG RAY_REPO=${HOME}/ray
-ARG RAY_COMMIT=448a83caf4
 ARG RAY_REPO_CACHE=/mnt/ray-cache
 ARG RAY_CACHE_CLEAR=false
+COPY --chown=${UID} build/ray-commit /tmp/ray-commit
```
glennmoy, line 125: Docker build can't find the `/build/ray-commit` file
https://github.com/beacon-biosignals/Ray.jl/actions/runs/6342266769/job/17227631104?pr=164#step:4:248

omus: Whoops, used a hypen here

glennmoy:
Shoudl we also checkout the same commit in `CI.yml`?

omus:
Initial Docker build is sooo slow: #165

glennmoy (approved):

PR #167, opened 2023-09-28
Fixes to build scripts for initial release

omus:
In attempting to do our initial release I encountered a few issues with things being out of date.

As I worked through these issues I noticed that the artifacts we are uploading have very specific triplets. Normally, the triplets from BinaryBuilder are based upon `BinaryBuilder.supported_platforms()` which uses quite generic triplets and mainly focuses on OS and architecture. Having an overly specific triplet is an issue in that it can result in us generating more artifacts than we have to. I've reduced the triplet down to what should be the current minimum. Ideally, we wouldn't use the Julia version in our triplet but appears that our `WORKSPACE.bazel` setup is pulling in Julia specific headers which is resulting in shared libraries which are incompatible with other Julia version (I think it's because our library is fully statically compiled).

Anyway, these should be the final changes needed to be merged so we can cut our initial release.

### DIFF
--- a/build/bind_artifacts.jl
+++ b/build/bind_artifacts.jl
@@ -7,8 +7,6 @@ using SHA: sha256
 using Tar
 using URIs: unescapeuri
 
-const DIR = mktempdir()
-
 include("common.jl")
 
 # Compute the Artifact.toml `git-tree-sha1`.
@@ -25,37 +23,23 @@ function sha256sum(tarball_path)
     end
 end
 
-function get_release_asset_urls()
-    io = IOBuffer()
-    Downloads.download(ASSETS_URL, io)
-    json = JSON3.read(String(take!(io)))
-    return unescapeuri.(get.(json[:assets], :browser_download_url))
-end
-
 if abspath(PROGRAM_FILE) == @__FILE__
-    @info "Fetching assets for $TAG"
-    artifacts_urls = get_release_asset_urls()
+    # Start with a clean Artifacts.toml so that unsupported platforms are removed
+    isfile(ARTIFACTS_TOML) && rm(ARTIFACTS_TOML)
 
-    for artifact_url in artifacts_urls
-        @info "Adding artifact for $(basename(artifact_url))"
-        artifact_path = joinpath(DIR, basename(artifact_url))
+    for platform in REQUIRED_PLATFORMS
+        artifact_name = gen_artifact_filename(; tag=TAG, platform)
+        artifact_url = gen_artifact_url(; repo_url=REPO_HTTPS_URL, tag=TAG, filename=artifact_name)
+
+        @info "Dowloading artifact $artifact_url"
+        artifact_path = joinpath(TARBALL_DIR, artifact_name)
         Downloads.download(artifact_url, artifact_path)
 
-        m = match(TARBALL_REGEX, basename(artifact_path))
-        if isnothing(m)
-            throw(ArgumentError("Could not parse host triplet from $(basename(artifact_path))"))
-        end
-
-        platform_triplet = m[:triplet]
-        julia_version = m[:julia_version]
-        platform_str = "$platform_triplet-julia_version+$julia_version"
-        platform = parse(BinaryPlatforms.Platform, platform_str)
-
-        bind_artifact!(JLL_ARTIFACTS_TOML,
+        @info "Adding artifact for $(triplet(platform))"
+        bind_artifact!(ARTIFACTS_TOML,
                        "ray_julia",
                        tree_hash_sha1(artifact_path);
                        platform=platform,
-                       download_info=[(artifact_url, sha256sum(artifact_path))],
-                       force=true)
+                       download_info=[(artifact_url, sha256sum(artifact_path))])
     end
 end
--- a/build/bind_artifacts.jl
+++ b/build/bind_artifacts.jl
@@ -29,7 +29,8 @@ if abspath(PROGRAM_FILE) == @__FILE__
 
     for platform in REQUIRED_PLATFORMS
         artifact_name = gen_artifact_filename(; tag=TAG, platform)
-        artifact_url = gen_artifact_url(; repo_url=REPO_HTTPS_URL, tag=TAG, filename=artifact_name)
+        artifact_url = gen_artifact_url(; repo_url=REPO_HTTPS_URL, tag=TAG,
+                                        filename=artifact_name)
 
         @info "Dowloading artifact $artifact_url"
         artifact_path = joinpath(TARBALL_DIR, artifact_name)
--- a/build/common.jl
+++ b/build/common.jl
@@ -23,7 +23,8 @@ const REQUIRED_JULIA_VERSIONS = (v"1.8", v"1.9")
 const REQUIRED_PLATFORMS = let
     base_platforms = parse.(Platform, REQUIRED_BASE_TRIPLETS)
     base_tags = [(; julia_version=string(v)) for v in REQUIRED_JULIA_VERSIONS]
-    [Platform(arch(p), platform_name(p); tags...) for p in base_platforms, tags in base_tags][:]
+    [Platform(arch(p), platform_name(p); tags...)
+     for p in base_platforms, tags in base_tags][:]
 end
 
 function remote_url(repo_root::AbstractString, name::AbstractString="origin")
--- a/build/common.jl
+++ b/build/common.jl
@@ -1,3 +1,4 @@
+using Base.BinaryPlatforms
 using LibGit2: LibGit2
 using Pkg.Types: read_project
 
@@ -14,8 +15,16 @@ const TARBALL_REGEX = r"""
 const GH_RELEASE_ASSET_PATH_REGEX = r"""
     ^/(?<owner>[^/]+)/(?<repo_name>[^/]+)/
     releases/download/
-    (?<tag>[^/]+)$
+    (?<tag>[^/]+)/?$
     """x
+
+const REQUIRED_BASE_TRIPLETS = ("x86_64-linux-gnu", "aarch64-apple-darwin")
+const REQUIRED_JULIA_VERSIONS = (v"1.8", v"1.9")
+const REQUIRED_PLATFORMS = let
+    base_platforms = parse.(Platform, REQUIRED_BASE_TRIPLETS)
+    base_tags = [(; julia_version=string(v)) for v in REQUIRED_JULIA_VERSIONS]
+    [Platform(arch(p), platform_name(p); tags...) for p in base_platforms, tags in base_tags][:]
+end
 
 function remote_url(repo_root::AbstractString, name::AbstractString="origin")
     return LibGit2.with(LibGit2.GitRepo(repo_root)) do repo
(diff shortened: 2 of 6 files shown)


build/bind_artifacts.jl
```diff
@@ -25,37 +23,24 @@ function sha256sum(tarball_path)
...
-    for artifact_url in artifacts_urls
-        @info "Adding artifact for $(basename(artifact_url))"
-        artifact_path = joinpath(DIR, basename(artifact_url))
-        Downloads.download(artifact_url, artifact_path)
-
-        m = match(TARBALL_REGEX, basename(artifact_path))
-        if isnothing(m)
-            throw(ArgumentError("Could not parse host triplet from $(basename(artifact_path))"))
-        end
+    for platform in REQUIRED_PLATFORMS
```
omus, line 30: By using a list of required platforms we can ensure that if we failed to upload all of the artifacts that the `bind_artifacts.jl` script will fail as it will attempt to download a file that doesn't exist

build/bind_artifacts.jl
```diff
@@ -25,37 +23,24 @@ function sha256sum(tarball_path)
...
+        artifact_name = gen_artifact_filename(; tag=TAG, platform)
+        artifact_url = gen_artifact_url(; repo_url=REPO_HTTPS_URL, tag=TAG,
+                                        filename=artifact_name)
 
-        platform_triplet = m[:triplet]
-        julia_version = m[:julia_version]
-        platform_str = "$platform_triplet-julia_version+$julia_version"
-        platform = parse(BinaryPlatforms.Platform, platform_str)
+        @info "Dowloading artifact $artifact_url"
+        artifact_path = joinpath(TARBALL_DIR, artifact_name)
```
omus, line 36: Previously we downloaded files to a temporary directory instead of the `TARBALL_DIR`. This will override any locally generated files but that seems fine.

build/common.jl
```diff
@@ -14,9 +15,18 @@ const TARBALL_REGEX = r"""
...
     """x
 
+const REQUIRED_BASE_TRIPLETS = ("x86_64-linux-gnu", "aarch64-apple-darwin")
+const REQUIRED_JULIA_VERSIONS = (v"1.8", v"1.9")
+const REQUIRED_PLATFORMS = let
+    base_platforms = parse.(Platform, REQUIRED_BASE_TRIPLETS)
+    base_tags = [(; julia_version=string(v)) for v in REQUIRED_JULIA_VERSIONS]
+    [Platform(arch(p), platform_name(p); tags...)
+     for p in base_platforms, tags in base_tags][:]
+end
```
omus, line 28: Didn't spend much time refactoring this. If we can make the artifacts non-Julia specific this becomes much cleaner

build/common.jl
```diff
@@ -25,17 +35,41 @@ function remote_url(repo_root::AbstractString, name::AbstractString="origin")
...
-const PKG_URL = remote_url(REPO_PATH)
+function set_url_scheme(url, scheme)
+    m = match(LibGit2.URL_REGEX, url)
+    if m === nothing
+        throw(ArgumentError("URL is not a valid SCP or HTTP(S) URL: $(url)"))
+    end
+    return LibGit2.git_url(; scheme="https", username=something(m[:user], ""),
+                           host=something(m[:host], ""), port=something(m[:port], ""),
+                           path=something(m[:path], ""))
+end
```
omus, line 46: @glennmoy I can understand your reason for copying the regex from LibGit2 as it's not exported by the library. However, now that I'm using `git_url` as well it seemed better to use the function and regex from the library that tests them.

build/common.jl
```diff
@@ -25,17 +35,41 @@ function remote_url(repo_root::AbstractString, name::AbstractString="origin")
...
 
-# Read Project.toml
-const JLL_PATH = joinpath(REPO_PATH, "ray_julia_jll")
-const JLL_PROJECT_TOML = joinpath(JLL_PATH, "Project.toml")
-const JLL_ARTIFACTS_TOML = joinpath(JLL_PATH, "Artifacts.toml")
+# Used to convert `HostPlatform` into something contained in
+# `BinaryBuilder.support_platforms()`
+function supported_platform(p::Platform)
+    support_tags = filter((k, v)::Pair -> k in ("call_abi", "libc"), tags(p))
+    support_tags["julia_version"] = string(Base.thisminor(VERSION))
```
omus, line 52: By using `Base.thisminor` we ensure that the `julia_version` is always `X.Y.0` which produces a more consistent artifact URL

build/upload_tarballs.jl
```diff
@@ -92,15 +64,9 @@ if abspath(PROGRAM_FILE) == @__FILE__
...
+    artifact_url = gen_artifact_url(; repo_url=REPO_HTTPS_URL, tag=TAG, filename="")
     branch = LibGit2.with(LibGit2.branch, LibGit2.GitRepo(REPO_PATH))
 
     @info "Uploading tarballs to $artifact_url"
-    try
-        upload_to_github_release(TARBALL_DIR, artifact_url, branch)
-    catch e
-        # ghr() already prints an error message with diagnosis
-        @debug "Caught exception $(sprint(showerror, e, catch_backtrace()))"
-    end
```
omus, line 105: The `try`/`catch` here not only suppressed `ghr` errors but also the `throw`n errors from the Julia code

omus:
@glennmoy feel free to approve and merge this when you get in. Once that's done we can make the `v0.0.1` release.

glennmoy:
>As I worked through these issues I noticed that the artifacts we are uploading have very specific triplets. Normally, the triplets from BinaryBuilder are based upon BinaryBuilder.supported_platforms() which uses quite generic triplets and mainly focuses on OS and architecture. Having an overly specific triplet is an issue in that it can result in us generating more artifacts than we have to.

Yeah I wasn't sure about having to include `libgfortran` etc since we're doing this outside BinaryBuilder...I went looking for alternative functions but concluded that perhaps `triplet()` was the canonical way to generate the artifact names? Seems it can be overly specific but maybe that's by design so the user can decide what not to include?

build/bind_artifacts.jl
```diff
@@ -25,37 +23,24 @@ function sha256sum(tarball_path)
...
+        Downloads.download(artifact_url, artifact_path)
```
glennmoy, line 37: will this fail if it tries to fetch an artifact that hasn't been uploaded yet?

build/common.jl
glennmoy: LGTM - it's not worth prettifying anyway

build/common.jl
glennmoy: So my understandig is this converts, e.g. `git@github.com:beacon-biosignals/ray.jl` to `https://github.com:beacon-biosignals/ray.jl`?

build/common.jl
```diff
@@ -25,17 +35,41 @@ function remote_url(repo_root::AbstractString, name::AbstractString="origin")
...
+    support_tags = [Symbol(k) => v for (k, v) in support_tags]
```
glennmoy, line 53: odd that the keys get converted to strings but the must be provided as kwargs...

build/common.jl
```diff
@@ -25,17 +35,41 @@ function remote_url(repo_root::AbstractString, name::AbstractString="origin")
...
-const TAG = "v$(JLL_PROJECT.version)"
+function gen_artifact_url(; repo_url, tag, filename)
+    return join([repo_url, "releases", "download", tag, filename], '/')
+end
 
-const GITHUB_URL = "https://api.github.com/repos"
-const _PKG = split(PKG_URL, ":")[2]
-const ASSETS_URL = joinpath(GITHUB_URL, "$_PKG", "releases", "tags", "$TAG")
+function gen_artifact_filename(; tag::AbstractString, platform::Platform)
+    return "ray_julia.$tag.$(triplet(platform)).tar.gz"
```
glennmoy, line 62:
```suggestion
    return "ray_julia.$tag.$(triplet(supported_platform(platform))).tar.gz"
```

I wonder do we want to call `supported_platform` again here for extra protection...although this function is only called over the `SUPPORTED_PLATFORMS` anyway it's probably overkill

build/upload_tarballs.jl
```diff
@@ -92,15 +64,9 @@ if abspath(PROGRAM_FILE) == @__FILE__
     # check that contents are tarballs with correct filename
     all(t -> !isnothing(match(TARBALL_REGEX, t)), readdir(TARBALL_DIR))
```
glennmoy, line 65: first time I wrote this I wanted to see if there was a `file` command (similar to the bash command) that would assert the file type rather than check against a regex. I went looking in `Base.Filesystem` module but didn't see anything. Only calling it out in case you knew of one off hand - not suggesting we have to change it now.

glennmoy (approved):
I'll give this a run through to make sure it works before merging

glennmoy:
Assuming this works we'll merge it, I'll regenerate the Artifacts and then we can make a tag!

build/bind_artifacts.jl
glennmoy: I checked and it does 👍

build/common.jl
omus: Yes, it converts SCP-like URLs to HTTPS URLs. Specifically `git@github.com:beacon-biosignals/Ray.jl` to `https://github.com/beacon-biosignals/Ray.jl` (there was a syntax issue with your HTTPS URL).

glennmoy:
@omus the build worked - I just have a few suggestions incoming

build/upload_tarballs.jl
omus: I think the usual way of handling this would be just to look at the extension. The other way is probably to use FileIO.jl and use the detection function which would detect the file type from magic bytes in the header: https://juliaio.github.io/FileIO.jl/stable/registry/#Registry-table

build/upload_tarballs.jl
```diff
@@ -92,15 +64,13 @@ if abspath(PROGRAM_FILE) == @__FILE__
...
 
-    pkg_http_url = parse_git_remote_url(PKG_URL)
-    artifact_url = "$(pkg_http_url)/releases/download/$TAG"
+    all(readdir(TARBALL_DIR)) do t
+        contains(t, "ray_julia.$TAG") || error("Stale tarball found: $t")
+    end
```
omus: Using `all` here is a bit strange due to having an internal error. A `for` or this seems better:
```suggestion
    stale = filter(!startswith("ray_julia.$TAG"), readdir("tarballs"))
    isempty(stale) || error("Stale tarball(s) found:\n$(join(stale, '\n'))")
```

build/upload_tarballs.jl
```diff
@@ -82,25 +47,30 @@ function upload_to_github_release(owner, repo_name, commit, tag, path;
...
 
 if abspath(PROGRAM_FILE) == @__FILE__
     isdir(TARBALL_DIR) || error("$TARBALL_DIR does not exist")
     !haskey(ENV, "GITHUB_TOKEN") && error("\"GITHUB_TOKEN\" environment variable required.")
 
     # check that contents are tarballs with correct filename
-    all(t -> !isnothing(match(TARBALL_REGEX, t)), readdir(TARBALL_DIR))
+    for t in readdir(TARBALL_DIR)
+        m = match(TARBALL_REGEX, t)
+        !isnothing(m) || error("Unexpected file found: tarballs/$t")
```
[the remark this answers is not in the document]
omus: Never mind. You're using the match groups later

PR #171, opened 2023-09-29
Add CI comment about tags ignoring paths

omus:
Follow up to #170. As it turns out when a workflow is triggered via a tag the `path` filter is ignored: https://github.com/orgs/community/discussions/26273.

The official documentation for this isn't clear at all and really it seems to be documented via omission. The only documentation on this I found was:

> If you define both `branches`/`branches-ignore` and `paths`/`paths-ignore`, the workflow will only run when both filters are satisfied

– https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-    tags: ["*"]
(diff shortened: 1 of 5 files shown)


glennmoy (approved):
~I'm guessing the `paths` still apply the `branches`?~ reread the PR description

PR #178, opened 2023-10-02
Build artifacts via GitHub Actions

omus:


omus:
Non-cached Bazel build timings:

|                                 | Build Ray CLI | Build ray_julia library |
|:--------------------------------|--------------:|------------------------:|
| Julia 1.8 - ubuntu-20.04-16core |       11m 34s |                  5m 24s |
| Julia 1.9 - ubuntu-20.04-16core |       10m 39s |                  5m 34s |
| Julia 1.8 - macos-latest-xlarge |       23m 15s |                 18m 33s |
| Julia 1.9 - macos-latest-xlarge |       22m 32s |                 18m 47s |

While monitoring these runs I noticed that the M1 runners (6-cores) were keeping up with the Linux runners (16-cores). The parallel difference did eventually make a difference but it appears that macOS has additional build steps.

omus:
Another round of timings:

|                                 | Build Ray CLI | Build ray_julia library |
|:--------------------------------|--------------:|------------------------:|
| Julia 1.8 - macos-latest-xlarge |       20m 44s |                  9m 0s  |
| Julia 1.9 - macos-latest-xlarge |       21m 44s |                  9m 20s |

Noticed that one of my M1 got stuck on the `julia-actions/setup-julia` action:
```
Request timeout: /bin/mac/aarch64/1.9/julia-1.9.3-macaarch64.tar.gz
Waiting 12 seconds before trying again
/usr/bin/tar xf /Users/runner/work/_temp/934c8e22-b774-4b77-ac2a-ac032d7ce895 --strip-components=1 -C julia-aarch64-1.9.3-zJoEGz
/Users/runner/hostedtoolcache/julia/1.9.3/aarch64/bin/julia --version
julia version 1.9.3
```

docs/src/building-artifacts.md
```diff
@@ -8,34 +8,57 @@ Follow [the instructions](https://github.com/beacon-biosignals/ray/blob/beacon-m
...
    julia --project upload_tarballs.jl
    ```
 
-5. Once all assets are uploaded, checkout a new branch (based off of `main`) and run the `bind_artifacts.jl`. This will modify the local `Artifacts.toml` with the artifact(s) associated with the Ray.jl version specified in the Project.toml. After running this you should commit and push the changes to a new PR.
+6. Run `bind_artifacts.jl` to modify local `Artifacts.toml` with the artifacts associated with the Ray.jl version specified in the `Project.toml`. After running this you should commit and push the changes to the PR you created in Step 1.
 
    ```sh
    julia --project bind_artifacts.jl
+
+   git commit -a -m "Update Artifacts.toml"
```
glennmoy, line 48:
```suggestion
   git commit Artifacts.toml -m "Update Artifacts.toml"
```

probabyl want to avoid accidental commits

build/build_tarballs.jl
```diff
@@ -13,16 +17,84 @@ function create_tarball(dir, tarball)
...
+    else
+        override = !("--no-override" in ARGS)
+        build_host_tarball(; override)
+    end
+
+    return nothing
+end
+
+if abspath(PROGRAM_FILE) == @__FILE__
+    main()
```
glennmoy, line 99: do we even use `main`? I feel like this is two functions stacked in a trenchcoat?

Might be better to pull out `download_ray_julia_artifacts` + friends into a `download_artifacts.jl` script.
That way we have a clean separation of what it means to build vs download and we don't have to worry about the mode-switching behaviour of `--fetch` (which kinda smells) and `--override` having no effect if fetching (another smell).

Conceptually I think it's much easier if `build_tarballs.jl --fetch` == `download_tarballs.jl` and is run locally in the build steps, while the new `build_tarballs.jl` is what runs in the CI.

glennmoy (approved):
Approving subject to pulling some stuff out of `build_tarballs.jl`

docs/src/building-artifacts.md
omus: Great idea! Made a minor adjustment:
```suggestion
   git commit -m "Update Artifacts.toml" -- Artifacts.toml
```

build/build_tarballs.jl
omus: I was worried about making too many scripts. However, if we did end up consolidating the scripts we'd probably combine a few more than what we have. Alright, I'll turn this into a new script.

PR #179, opened 2023-10-02
Enable Apple silicon CI runners

omus:
Using the newly announced M1 GHA runners: https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/. We should test against M1 in CI as it's one of the very limited platforms we support for Ray.jl and doing this should unlock the ability for automated artifact builds for making new releases.

I did some initial non-cache timings as part of another PR: https://github.com/beacon-biosignals/Ray.jl/pull/178#issuecomment-1743567864

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -58,6 +58,11 @@ jobs:
           repository: beacon-biosignals/ray
           path: ${{ env.ray_dir }}
           ref: ${{ steps.ray-commit.outputs.sha }}
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: pip
+          cache-dependency-path: ${{ env.ray_dir }}/python/setup.py
       - uses: julia-actions/setup-julia@v1
         with:
           version: ${{ matrix.version }}
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -137,9 +137,11 @@ jobs:
           popd
 
           # By copying the entire Ray worktree we can easily restore missing files without having to
-          # delete the cache and build from scratch.
-          mkdir -p "$RAY_GEN_CACHE_DIR"
-          cp -rfp . "$RAY_GEN_CACHE_DIR"
+          # delete the cache and build from scratch. Skip copy when we don't save the cache.
+          if [ "${{ steps.build-cache.outputs.cache-hit }}" != "true" ]; then
+              mkdir -p "$RAY_GEN_CACHE_DIR"
+              cp -rfp . "$RAY_GEN_CACHE_DIR"
+          fi
 
           # Verify Ray CLI works
           echo "Verify Ray CLI works" >&2
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -84,7 +84,6 @@ jobs:
       # https://docs.ray.io/en/releases-2.5.1/ray-contribute/development.html#building-ray-on-linux-macos-full
       - name: Build Ray CLI
         run: |
-          set -eux
           pip install --upgrade pip wheel
 
           # The Ray BUILD.bazel includes a bunch of `copy_to_workspace` rules which copy build output
@@ -94,7 +93,6 @@ jobs:
           RAY_GEN_CACHE_DIR=~/.cache/ray-generated
           if [ -d "$RAY_GEN_CACHE_DIR" ]; then
               dest_root=$(pwd)
-              pushd $RAY_GEN_CACHE_DIR
               rel_srcs=(
                   python/ray/_raylet.so \
                   python/ray/core/generated \
@@ -103,12 +101,16 @@ jobs:
                   python/ray/core/src/ray/gcs \
               )
               for rel in "${rel_srcs[@]}"; do
+                  echo "Restoring $rel" >&2
+                  src="$RAY_GEN_CACHE_DIR/$rel"
                   dest="$dest_root/$rel"
-                  echo "Restoring $rel"
                   mkdir -p $(dirname $dest)
-                  cp -rp $rel $dest
+                  cp -rp $src $dest
+                  echo "src:"
+                  ls -l $src
+                  echo "dest:"
+                  ls -l $dest
               done
-              popd
           fi
 
           pushd python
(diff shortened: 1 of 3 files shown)


.github/workflows/artifacts_CI.yml
```diff
@@ -20,18 +20,24 @@ concurrency:
...
     strategy:
       fail-fast: false
       matrix:
         version:
           - "1.8"  # Version used with MVP
           - "1.9"  # Latest release
         os:
-          - ubuntu-latest
+          - ubuntu-20.04-16core
+          - macos-latest-xlarge  # M1
```
omus, line 32: Enabling the M1 runner in the artifacts CI ensures that the M1 artifacts are properly updated and not broken. This workflow usually only runs when the `Artifacts.toml` is changed so this is critical for ensuring a potential release is working.

omus:
I need to find a macOS alternative to `cp --parents`:
```
 ~/.cache/ray-generated ~/work/Ray.jl/Ray.jl/build/ray
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-aclpsvXx] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-aclpsvXx] source_file ... target_directory
```

omus:
Going to wait until the Julia CI jobs complete before pushing any other changes as that will let the GHA cache populate and have subsequent runs go much faster.

omus:
TODOs for tomorrow:
- [ ] Determine why Linux CI runners are failing with `ModuleNotFoundError: No module named 'ray.core.generated.common_pb2'`
- [ ] Determine why CI caching isn't working properly on Apple silicon (more rebuilding is happening that I would expect)

omus:
Clearing the cache to see if a clean start helps

.github/workflows/CI.yml
```diff
@@ -86,15 +92,21 @@ jobs:
...
                   python/ray/_raylet.so \
                   python/ray/core/generated \
                   python/ray/serve/generated \
                   python/ray/core/src/ray/raylet/raylet \
                   python/ray/core/src/ray/gcs \
-                  $dest
+              )
+              for rel in "${rel_srcs[@]}"; do
+                  dest="$dest_root/$rel"
+                  echo "Restoring $rel -> $dest"
```
omus, line 106:
```suggestion
                  echo "Restoring $rel"
```

omus:
Ah, this may be the issue with the cache:
```
/opt/homebrew/bin/gtar: ../../../.cache/ray-generated/bazel-ray/external/local_config_cc/cc_toolchain_config.bzl: time stamp 2033-09-30 05:44:55 is 315329196.733191 s in the future
Cache restored successfully
```

omus:
Fresh cache didn't change anything

omus:
> Determine why Linux CI runners are failing with ModuleNotFoundError: No module named 'ray.core.generated.common_pb2'

Issue was that the re-implementation of `cp --parents` wasn't copying to the correct destination resulting in the Ray CLI build succeeding but the resulting CLI being broken at runtime. I've added a `ray --version` self-test to ensure that we catch these this kind of failure during the build step and not later when tests attempt to use the Ray CLI.

omus:
With the previous commit I fixed an issue where the `hashFiles` paths weren't updated when we did #158. With that change the cache key should now be correct. All of the CI workflow jobs will generate fresh cache entries which on the macOS side should also include the Bazel output base files which we need to cache.

omus:
When listing the contents on macOS without a CI cache and before building I see:
```
-----------
find -L ~/.cache
/Users/runner/.cache
/Users/runner/.cache/powershell
/Users/runner/.cache/powershell/ModuleAnalysisCache-D0C7B8E2
/Users/runner/.cache/powershell/PowerShellGet
/Users/runner/.cache/powershell/PowerShellGet/PSRepositories.xml
/Users/runner/.cache/powershell/StartupProfileData-NonInteractive
/Users/runner/.cache/powershell/telemetry.uuid
-----------
find -L /private/var/tmp
/private/var/tmp
/private/var/tmp/kernel_panics
/private/var/tmp/_bazel_runner
/private/var/tmp/_bazel_runner/install
/private/var/tmp/_bazel_runner/install/d1c493495cc9599d12332f6800b1abcf
/private/var/tmp/_bazel_runner/install/d1c493495cc9599d12332f6800b1abcf/build-runfiles
...
```
It's interesting that there already is some Bazel cache information present on the machine before we use Bazel. At the moment I'm deleting this information to avoid bloating our cache. If it was helping our build times I would consider keeping it but from what I've seen so far it isn't used at all.

omus:
Might be able to utilize this in the future to narrow down what we cache:
```
❯ bazel info output_base
/private/var/tmp/_bazel_cvogt/8c1a2d6870327d0b41c86ac2d10e8f99
```

omus:
The Python cache seemed to have done the trick. Now the macOS "Build Ray CLI" is 42s

omus:
CI / Julia workflow timings:

|                                            | Restore build cache | Build Ray CLI | Build ray_julia library | julia-runtests |
|:-------------------------------------------|--------------------:|--------------:|------------------------:|---------------:|
| Julia 1.8 - ubuntu-20.04-16core (no cache) |    0s | 12m 34s | 5m 10s | 2m 58s |  
| Julia 1.9 - ubuntu-20.04-16core (no cache) |    0s | 12m  9s | 5m 16s | 2m 50s |
| Julia 1.8 - macos-latest-xlarge (no cache) |    1s | 20m 28s | 8m 44s | 2m  0s |
| Julia 1.9 - macos-latest-xlarge (no cache) |    0s | 21m 26s | 9m 21s |      - |
| Julia 1.8 - ubuntu-20.04-16core (cached)   |   19s |     23s |    22s | 2m 58s |
| Julia 1.9 - ubuntu-20.04-16core (cached)   |   19s |     45s |    23s | 2m 40s |
| Julia 1.8 - macos-latest-xlarge (cached)   | 2m 2s |     42s | 1m 10s | 1m 59s |
| Julia 1.9 - macos-latest-xlarge (cached)   | 2m 0s |     41s |    49s | 1m 50s |

– https://github.com/beacon-biosignals/Ray.jl/actions/runs/6397292429
- https://github.com/beacon-biosignals/Ray.jl/actions/runs/6396789493

kleinschmidt (approved):
too bad the cache restore times are so long, but still a win over un-cached, and very good to be able to build our apple silicon JLL binaries in CI!

PR #182, opened 2023-10-03
use reviewdog suggestions for format, only on non-draft PRs

kleinschmidt:
Confirmed that
- will ~~not~~ format draft PRs
- format errors leave comments with suggestion to fix
- ...except on draft PRs, where the build fails

### DIFF
--- a/.github/workflows/FormatCheck.yml
+++ b/.github/workflows/FormatCheck.yml
@@ -43,6 +43,7 @@ jobs:
           using Pkg
           Pkg.add(PackageSpec(name="JuliaFormatter", version="1"))
       - name: Check formatting
+        id: check
         shell: julia --color=yes {0}
         run: |
           using JuliaFormatter
--- a/.github/workflows/FormatCheck.yml
+++ b/.github/workflows/FormatCheck.yml
@@ -18,6 +18,7 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   format-check:
+    if: (github.event_name == 'push') || (github.event.pull_request.draft == false)
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -58,5 +59,9 @@ jobs:
           end
           if !formatted
               @error "Found some unformatted files:\n" * read(`git diff --name-only`, String)
-              exit(1)
           end
+      - uses: reviewdog/action-suggester@v1
+        if: github.event_name == 'pull_request'
+        with:
+          tool_name: JuliaFormatter
+          fail_on_error: true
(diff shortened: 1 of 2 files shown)


.github/workflows/FormatCheck.yml
```diff
@@ -18,6 +18,7 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   format-check:
+    if: (github.event_name == 'push') || (github.event.pull_request.draft == false)
```
omus, line 21:
```suggestion
    if: github.event_name == 'push' || github.event.pull_request.draft == false
```

.github/workflows/FormatCheck.yml
```diff
@@ -58,5 +59,9 @@ jobs:
           end
           if !formatted
               @error "Found some unformatted files:\n" * read(`git diff --name-only`, String)
-              exit(1)
           end
+      - uses: reviewdog/action-suggester@v1
+        if: github.event_name == 'pull_request'
+        with:
+          tool_name: JuliaFormatter
+          fail_on_error: true
```
omus, line 67: It would be nice if the format workflow would run all the time but just avoid doing suggestions until out of draft mode. This could look something like this:
```yaml
jobs:
  format-check:
    env:
      add_suggestions: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
    steps:
      - name: Check formatting
        run: |
          ...
          if !formatted
              @error "Found some unformatted files:\n" * read(`git diff --name-only`, String)
              get(ENV, "add_suggestions", "false") == "false" && exit(1)
          end
      - uses: reviewdog/action-suggester@v1
        if: ${{ env.add_suggestions }}
        with:
          tool_name: JuliaFormatter
          fail_on_error: true
```

omus: The problem with the current setup is that a formatting error to `main` won't cause a workflow failure

omus: Another option I like better:
```yaml
jobs:
  format-check:
    steps:
      - name: Check formatting
        id: check
        run: |
          ...
          if !formatted
              @error "Found some unformatted files:\n" * read(`git diff --name-only`, String)
          end
          open(ENV["GITHUB_OUTPUT"], "a") do io
              println(io, "formatted=$formatted")
          end
      - uses: reviewdog/action-suggester@v1
        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
        with:
          tool_name: JuliaFormatter
          fail_on_error: true
      - name: Fail if not formatted
        if: ${{ steps.check.outputs.formatted != 'true' }}
        run: exit 1
```

omus (approved):
Fantastic

PR #187, opened 2023-10-06
Fix `log_deserialization_error` bug

omus:
Fixes a bug introduced in #140.


kleinschmidt (approved):
if we're hitting this path in actual practice, is there some way to hit it in the tests for `deserialize_from_ray_object`?

omus:
> if we're hitting this path in actual practice, is there some way to hit it in the tests for `deserialize_from_ray_object`?

Good point. I think I can cook up a test for that. I may just drop the new testset then

PR #190, opened 2023-10-06
Return Status from CoreWorker function and remove RAY_CHECK

### LINKED ISSUE #84: Avoid using `RAY_CHECK*`
> We shouldn't be using `RAY_CHECK*` because when it fails, it exits the process.  better to check + throw.  but we can handle that in a cleanup later.

_Originally posted by @kleinschmidt in https://github.com/beacon-biosignals/ray_core_worker_julia_jll.jl/pull/83#discussion_r1317280854_

glennmoy:
Closes #84

### DIFF
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -110,13 +110,17 @@ void initialize_worker(
                         application_error,
                         is_retryable_error);
 
-          RAY_CHECK(return_vec.size() == 1);
-
           // TODO: support multiple return values
           // https://github.com/beacon-biosignals/Ray.jl/issues/54
-          (*returns)[0].second = return_vec[0];
-          return Status::OK();
+          if (return_vec.size() == 1) {
+            (*returns)[0].second = return_vec[0];
+            return Status::OK();
+          } 
+          else {
+            return Status::Invalid("Multiple values returned by task");
+          };
         };
+
     RAY_LOG(DEBUG) << "ray_julia_jll: Initializing julia worker coreworker";
     CoreWorkerProcess::Initialize(options);
 
@@ -187,7 +191,10 @@ ObjectID put(const std::shared_ptr<RayObject> object,
 
     // Store our data in the object store
     ObjectID object_id;
-    RAY_CHECK_OK(worker.Put(*object, contained_object_ids, &object_id));
+    auto status = worker.Put(*object, contained_object_ids, &object_id);
+    if (!status.ok()) {
+        throw std::runtime_error(status.ToString());
+    }
     return object_id;
 }
 
@@ -203,7 +210,11 @@ std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
         return nullptr;
     }
 
-    RAY_CHECK(objects.size() == 1);
+    if (objects.size() != 1) {
+        auto st = Status::Invalid("Size of RayObject is larger than 1");
+        throw std::runtime_error(st.ToString());
+    }
+
     return objects[0];
 }
 
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -195,16 +195,18 @@ Status put(const std::shared_ptr<RayObject> object,
 }
 
 // Example of using `CoreWorker::Get`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L210-L220
-Status get(const ObjectID object_id, const int64_t timeout_ms, std::vector<std::shared_ptr<RayObject>> *objects) {
+Status get(const ObjectID object_id, const int64_t timeout_ms, std::shared_ptr<RayObject> *result) {
     auto &worker = CoreWorkerProcess::GetCoreWorker();
 
     // Retrieve our data from the object store
     std::vector<ObjectID> get_obj_ids = {object_id};
-    auto status = worker.Get(get_obj_ids, timeout_ms, objects);
+    std::vector<shared_ptr<RayObject>> result_vec;
+    auto status = worker.Get(get_obj_ids, timeout_ms, &result_vec);
+    *result = result_vec[0];
 
-    // TODO (maybe?): allow multiple return values
-    // https://github.com/beacon-biosignals/Ray.jl/issues/54
-    auto num_objs = objects->size();
+     // TODO (maybe?): allow multiple return values
+     // https://github.com/beacon-biosignals/Ray.jl/issues/54
+    auto num_objs = result_vec.size();
     if (num_objs != 1) {
         auto msg = "Requested a single object but instead found " + std::to_string(num_objs) + " objects.";
         status = Status::UnknownError(msg);
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -207,7 +207,7 @@ std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
     std::vector<ObjectID> get_obj_ids = {object_id};
     auto status = worker.Get(get_obj_ids, timeout_ms, &objects);
     if (!status.ok()) {
-        return nullptr;
+        throw std::runtime_error(status.ToString());
     }
 
     if (objects.size() != 1) {
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -203,11 +203,13 @@ Status get(const ObjectID object_id, int64_t timeout_ms, std::vector<std::shared
     std::vector<ObjectID> get_obj_ids = {object_id};
     auto status = worker.Get(get_obj_ids, timeout_ms, objects);
 
-    // if (objects.size() != 1) {
-    //     auto msg = "Requested a single object but instead found " + std::to_string(objects.size()) + " objects.";
-    //     auto st = Status::UnknownError(msg);
-    //     return st;
-    // }
+    // TODO: allow multiple return values
+    auto num_objs = objects->size();
+    if (num_objs != 1) {
+        auto msg = "Requested a single object but instead found " + std::to_string(num_objs) + " objects.";
+        auto st = Status::UnknownError(msg);
+        return st;
+    }
 
     return status;
 }
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -117,8 +117,8 @@ void initialize_worker(
             return Status::OK();
           } 
           else {
-            auto msg = "Task returned too many values: " + std::to_string(return_vec.size());
-            return Status::Invalid(msg);
+            auto msg = "Task returned " + std::to_string(return_vec.size()) + " values. Expected 1.";
+            return Status::NotImplemented(msg);
           };
         };
 
@@ -212,8 +212,8 @@ std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
     }
 
     if (objects.size() != 1) {
-        auto msg = "Size of RayObjects is " + std::to_string(objects.size());
-        auto st = Status::Invalid(msg);
+        auto msg = "Requested a single object but instead found " + std::to_string(objects.size()) + " objects.";
+        auto st = Status::UnknownError(msg);
         throw std::runtime_error(st.ToString());
     }
 
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -115,7 +115,7 @@ void initialize_worker(
           if (return_vec.size() == 1) {
             (*returns)[0].second = return_vec[0];
             return Status::OK();
-          } 
+          }
           else {
             auto msg = "Task returned " + std::to_string(return_vec.size()) + " values. Expected 1.";
             return Status::NotImplemented(msg);
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -117,7 +117,8 @@ void initialize_worker(
             return Status::OK();
           } 
           else {
-            return Status::Invalid("Multiple values returned by task");
+            auto msg = "Task returned too many values: " + std::to_string(return_vec.size());
+            return Status::Invalid(msg);
           };
         };
 
@@ -211,7 +212,8 @@ std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
     }
 
     if (objects.size() != 1) {
-        auto st = Status::Invalid("Size of RayObject is larger than 1");
+        auto msg = "Size of RayObjects is " + std::to_string(objects.size());
+        auto st = Status::Invalid(msg);
         throw std::runtime_error(st.ToString());
     }
 
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -195,7 +195,7 @@ Status put(const std::shared_ptr<RayObject> object,
 }
 
 // Example of using `CoreWorker::Get`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L210-L220
-Status get(const ObjectID object_id, int64_t timeout_ms, std::vector<std::shared_ptr<RayObject>> *objects) {
+Status get(const ObjectID object_id, const int64_t timeout_ms, std::vector<std::shared_ptr<RayObject>> *objects) {
     auto &worker = CoreWorkerProcess::GetCoreWorker();
 
     // Retrieve our data from the object store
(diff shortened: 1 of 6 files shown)


build/wrapper.cc
```diff
@@ -110,13 +110,18 @@ void initialize_worker(
...
           // https://github.com/beacon-biosignals/Ray.jl/issues/54
-          (*returns)[0].second = return_vec[0];
-          return Status::OK();
+          if (return_vec.size() == 1) {
+            (*returns)[0].second = return_vec[0];
+            return Status::OK();
+          } 
+          else {
+            auto msg = "Task returned too many values: " + std::to_string(return_vec.size());
+            return Status::Invalid(msg);
```
glennmoy, line 121: unsure what type of `Status` we want to return here? I went with `Invalid` as it seemed the most generic

omus: `NotImplemented` seems appropriate

build/wrapper.cc
```diff
@@ -187,7 +192,10 @@ ObjectID put(const std::shared_ptr<RayObject> object,
 
     // Store our data in the object store
     ObjectID object_id;
-    RAY_CHECK_OK(worker.Put(*object, contained_object_ids, &object_id));
+    auto status = worker.Put(*object, contained_object_ids, &object_id);
+    if (!status.ok()) {
+        throw std::runtime_error(status.ToString());
+    }
     return object_id;
```
omus, line 199: So the actual [`CoreWorker::Put`](https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/core_worker/core_worker.h#L558-L560) returns just `Status` and passes back the `object_id` by reference. I'd try that approach here.

build/wrapper.cc
```diff
@@ -200,10 +208,15 @@ std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
...
+    }
+
+    if (objects.size() != 1) {
+        auto msg = "Size of RayObjects is " + std::to_string(objects.size());
+        auto st = Status::Invalid(msg);
+        throw std::runtime_error(st.ToString());
     }
 
-    RAY_CHECK(objects.size() == 1);
     return objects[0];
```
omus, line 220: Similarly to `CoreWorker::Put` we should have [`CoreWorker::Get`](https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/core_worker/core_worker.h#L665-L667) return a `Status` and pass back the object by reference.

build/wrapper.cc
```diff
@@ -200,10 +208,15 @@ std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
...
     if (!status.ok()) {
-        return nullptr;
+        throw std::runtime_error(status.ToString());
+    }
+
+    if (objects.size() != 1) {
+        auto msg = "Size of RayObjects is " + std::to_string(objects.size());
+        auto st = Status::Invalid(msg);
+        throw std::runtime_error(st.ToString());
     }
```
omus, line 218: This is an internal error as the caller can only specify one object to get. As there is no "InternalError" I think "UnknownError" probably fits best:
```suggestion
    if (objects.size() != 1) {
        auto msg = "Requested a single object but instead found " + std::to_string(objects.size()) + " objects.";
        auto st = Status::UnknownError(msg);
        throw std::runtime_error(st.ToString());
    }
```

build/wrapper.cc
omus: It's definitely a little bit of scope creep but it should clean things up nicely

build/wrapper.cc
```diff
@@ -110,13 +110,18 @@ void initialize_worker(
...
-          (*returns)[0].second = return_vec[0];
-          return Status::OK();
+          if (return_vec.size() == 1) {
+            (*returns)[0].second = return_vec[0];
+            return Status::OK();
+          } 
+          else {
+            auto msg = "Task returned " + std::to_string(return_vec.size()) + " values. Expected 1.";
+            return Status::NotImplemented(msg);
+          };
```
glennmoy, line 122: I don't think there's any need for this restriction anymore... I'll make a quick attempt to obviate it

glennmoy: moving this into https://github.com/beacon-biosignals/Ray.jl/pull/193 as it contains more changes than I think is suitable for this PR

src/object_ref.jl
```diff
@@ -63,7 +63,8 @@ function Base.deepcopy_internal(x::ObjectRef, stackdict::IdDict)
     return xcp
 end
 
-function ObjectRef(oid::ray_jll.ObjectIDAllocated; kwargs...)
+function ObjectRef(oid::Union{ray_jll.ObjectIDAllocated,ray_jll.ObjectIDDereferenced};
+                   kwargs...)
```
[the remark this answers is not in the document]
kleinschmidt: that's just 
```suggestion
function ObjectRef(oid::ray_jll.ObjectID; kwargs...)
```

right?

build/wrapper.cc
```diff
@@ -180,31 +185,33 @@ ray::core::CoreWorker &_GetCoreWorker() {
...
-    ObjectID object_id;
-    RAY_CHECK_OK(worker.Put(*object, contained_object_ids, &object_id));
-    return object_id;
+    auto status = worker.Put(*object, contained_object_ids, object_id);
+    return status;
 }
 
 // Example of using `CoreWorker::Get`: https://github.com/ray-project/ray/blob/ray-2.5.1/src/ray/core_worker/test/core_worker_test.cc#L210-L220
-std::shared_ptr<RayObject> get(const ObjectID object_id, int64_t timeout_ms) {
+Status get(const ObjectID object_id, int64_t timeout_ms, std::vector<std::shared_ptr<RayObject>> *objects) {
```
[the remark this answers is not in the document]
kleinschmidt: (I agree with @omus suggestion here though)

kleinschmidt (commented):
this is a bit more in-depth than what I had in mind, which was to throw runtime errors instead of using the `RAY_CHECK` macro.  but generally in the right direction.

if we're going to make all these JLL functions have the same kind of interface as the C++ ones (pass in pointer to return args, return `Status`), I think there are a few more places that need to be brought into alignment with that (look for places where the status is checked for `ok()` and runtime exceptiosn raised if not...)

build/wrapper.cc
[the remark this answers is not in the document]
glennmoy: So..having trouble with the "re-instante the vector" part...

We give a pointer to a ` shared_ptr<RayObject>` (`result`), but the coreworker function takes a pointer to a `vector<shared_ptr<RayObject>>` (`result_vec`).

It's not as simple as wrappring `result` in a vector since then we'd have a pointer to a vector of _pointers_.
And it's also not viable to set the `result` to point to `&result_vec[0]` after the fact either since I'm pretty sure that memory is released once `get` returns.

I tried to `memcpy` the actual data in `result_vec` into the address of `result` but that wasn't working out either.

I want to completely avoid `reinterpret_cast` and friends since those are too dangerous but I'm open to any ideas?

IMO, I think passing a `vector<shared_ptr<RayObject>>` would be fine as  it's consistent with the function signature of the coreworker functions. It's also much easier to extract the `only` value in the `results` on the julia-side.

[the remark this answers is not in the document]
glennmoy: > I also tried to memcpy the actual data in result_vec into the address of result but that wasn't working out either.
Specifically I used `std::memcpy(result, &result_vec[0], sizeof(shared_ptr<RayObject>))`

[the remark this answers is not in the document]
kleinschmidt: > In any case IMO, I think passing a vector<shared_ptr<RayObject>> would be fine as it's consistent with the function signature of the coreworker functions. It's also much easier to extract the only value in the results on the julia-side.

Okay, that's fine.  Don't get too bogged down here.  I'm annoyed that we are leaking so much of the C++ stuff into the Ray.jl codebase but it seems like that ship has sailed, especially since we've lost the distinction between the jll packager and Ray.jl itself.

Just leave a comment in the julia code explaining why you have to wrap in a vector.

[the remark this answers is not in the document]
omus: > And it's also not viable to set the `result` to point to `&result_vec[0]` after the fact either since I'm pretty sure that memory is released once get returns.

As we're dealing with `shared_ptr`s here as long as you have one pointer to the memory you are using it will be retained. We should definitely be able to do this but I suspect you'll need to have a `std::shared_ptr<RayObject> *` so you can update the passed in parameter. I'll note this [`std::swap`](https://cplusplus.com/reference/memory/shared_ptr/swap-free/) function kinda does what we want but I don't think it's necessary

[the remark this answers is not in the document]
omus: This is my impression on how this would be done:
```cpp
Status get(const ObjectID object_id, int64_t timeout_ms, std::shared_ptr<RayObject> *result) {
    std::vector<std::shared_ptr<RayObject>> objects;
    auto status = worker.Get(get_obj_ids, timeout_ms, &objects);
    *result = objects[0];
    ...
}
```
The `std::shared_ptr`'s data is stored on the heap.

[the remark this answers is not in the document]
glennmoy: I will give this a shot..I had initially tried `result = &objects[0]` but I realise in hindsight I was just reassigning the _address_ of `result` and not the contents.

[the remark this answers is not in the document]
glennmoy: worked 👍

build/wrapper.cc
```diff
@@ -180,31 +185,34 @@ ray::core::CoreWorker &_GetCoreWorker() {
...
 
     // Retrieve our data from the object store
-    std::vector<std::shared_ptr<RayObject>> objects;
     std::vector<ObjectID> get_obj_ids = {object_id};
-    auto status = worker.Get(get_obj_ids, timeout_ms, &objects);
-    if (!status.ok()) {
-        return nullptr;
+    std::vector<shared_ptr<RayObject>> result_vec;
+    auto status = worker.Get(get_obj_ids, timeout_ms, &result_vec);
+    *result = result_vec[0];
```
glennmoy, line 205: this was infuriatingly simple in the end 🤦

omus: Thanks for sticking with it ❤️

src/object_store.jl
```diff
@@ -26,9 +28,11 @@ captured exception will be thrown on `get`.
...
 function get(obj_ref::ObjectRef)
     wait(obj_ref)
-    ray_obj = ray_jll.get(obj_ref.oid, 0)
-    isnull(ray_obj[]) && error("got null pointer after successful `wait`; this is a bug!")
-    return deserialize_from_ray_object(ray_obj, obj_ref)
+    ray_objs = CxxPtr(SharedPtr{ray_jll.RayObject}())
+    status = ray_jll.get(obj_ref.oid, 0, ray_objs)
+    Symbol(status) == :OK || error("ray_julia_jll.get returned Status::$status")
+    isnull(ray_objs) && error("got null pointer after successful `wait`; this is a bug!")
+    return deserialize_from_ray_object(ray_objs[], obj_ref)
```
omus, line 35:
```suggestion
    ray_obj = CxxPtr(SharedPtr{ray_jll.RayObject}())
    status = ray_jll.get(obj_ref.oid, 0, ray_obj)
    Symbol(status) == :OK || error("ray_julia_jll.get returned Status::$status")
    isnull(ray_obj) && error("got null pointer after successful `wait`; this is a bug!")
    return deserialize_from_ray_object(ray_obj[], obj_ref)
```

test/ray_julia_jll/put_get.jl
```diff
@@ -11,13 +11,17 @@ using .ray_julia_jll: ObjectID
...
+        @test Symbol(status) == :OK
+        @test ray_julia_jll.contains(oid[])
 
         # TODO: Currently uses size/length from `data`
         # https://github.com/beacon-biosignals/Ray.jl/issues/55
-        ray_obj = ray_julia_jll.get(oid, -1)
-        buffer = GetData(ray_obj[])
+        ray_obj = CxxPtr(SharedPtr{ray_jll.RayObject}())
+        status = ray_julia_jll.get(oid[], -1, ray_obj)
+        @test Symbol(status) == :OK
```
omus, line 23: Testing it this way ensures that a `Status` type is returned or at least comparable
```suggestion
        @test status == Status(:OK)
```
Will need to import `Status`

test/ray_julia_jll/put_get.jl
```diff
@@ -32,11 +36,15 @@ using .ray_julia_jll: ObjectID
     @testset "roundtrip string" begin
         data = "Greetings from Julia!"
         buffer = LocalMemoryBuffer(Ptr{Nothing}(pointer(data)), sizeof(data), true)
-        oid = ray_julia_jll.put(RayObject(buffer), StdVector{ObjectID}())
-        @test ray_julia_jll.contains(oid)
+        oid = CxxPtr(ray_jll.ObjectID())
+        status = ray_julia_jll.put(RayObject(buffer), StdVector{ObjectID}(), oid)
+        @test Symbol(status) == :OK
```
omus, line 41:
```suggestion
        @test status == Status(:OK)
```

test/ray_julia_jll/put_get.jl
```diff
@@ -32,11 +36,15 @@ using .ray_julia_jll: ObjectID
...
+        @test ray_julia_jll.contains(oid[])
 
-        ray_obj = ray_julia_jll.get(oid, -1)
-        buffer = GetData(ray_obj[])
+        ray_obj = CxxPtr(SharedPtr{ray_jll.RayObject}())
+        status = ray_julia_jll.get(oid[], -1, ray_obj)
+        @test Symbol(status) == :OK
```
omus, line 46:
```suggestion
        @test status == Status(:OK)
```

test/ray_julia_jll/put_get.jl
```diff
@@ -11,13 +11,17 @@ using .ray_julia_jll: ObjectID
     @testset "roundtrip vector" begin
         data = UInt16[1:3;]
         buffer = LocalMemoryBuffer(Ptr{Nothing}(pointer(data)), sizeof(data), true)
-        oid = ray_julia_jll.put(RayObject(buffer), StdVector{ObjectID}())
-        @test ray_julia_jll.contains(oid)
+        oid = CxxPtr(ray_jll.ObjectID())
+        status = ray_julia_jll.put(RayObject(buffer), StdVector{ObjectID}(), oid)
+        @test Symbol(status) == :OK
```
omus, line 16:
```suggestion
        @test status == Status(:OK)
```

build/wrapper.cc
```diff
@@ -110,13 +110,18 @@ void initialize_worker(
...
-          (*returns)[0].second = return_vec[0];
-          return Status::OK();
+          if (return_vec.size() == 1) {
+            (*returns)[0].second = return_vec[0];
+            return Status::OK();
+          }
+          else {
+            auto msg = "Task returned " + std::to_string(return_vec.size()) + " values. Expected 1.";
+            return Status::NotImplemented(msg);
+          };
```
omus, line 122: Indentation
```suggestion
          if (return_vec.size() == 1) {
              (*returns)[0].second = return_vec[0];
              return Status::OK();
          }
          else {
              auto msg = "Task returned " + std::to_string(return_vec.size()) + " values. Expected 1.";
              return Status::NotImplemented(msg);
          };
```

omus (approved):
A couple minor things

test/ray_julia_jll/put_get.jl
glennmoy: ah yeah I was trying to figure out the inverse of that and was using `StatusCode` 👍

build/wrapper.cc
glennmoy: the whole block is indented wrong

test/ray_julia_jll/put_get.jl
glennmoy: I think the difference is that the Enum is a `StatusCode` while the return of the function is a `Status`

glennmoy: alt:
```julia
julia> ray_julia_jll.ok(status)
true

julia> ray_julia_jll.code(status) == ray_julia_jll.StatusCode(:OK)
true
```

Issue #54, opened 2023-08-29
Support multiple return values from task exectutor


glennmoy:

omus:
We may not actually want this. Until we find a good use case for this in Julia it makes sense to keep only returning a single `ObjectRef` from Julia Ray tasks.

PR #193, opened 2023-10-10
Allow Ray task to return multiple variables

glennmoy:
closes https://github.com/beacon-biosignals/Ray.jl/issues/54

### DIFF
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -216,7 +216,7 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
         ""
     end
 
-    oid = GC.@preserve task_args begin
+    oids = GC.@preserve task_args begin
         ray_jll._submit_task(fd,
                              transform_task_args(task_args),
                              serialized_runtime_env_info,
@@ -224,7 +224,7 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
     end
     # CoreWorker::SubmitTask calls TaskManager::AddPendingTask which initializes
     # the local ref count to 1, so we don't need to do that here.
-    return ObjectRef(oid; add_local_ref=false)
+    return ObjectRef(only(oids); add_local_ref=false)
 end
 
 # Adapted from `prepare_args_internal`:
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -110,16 +110,10 @@ void initialize_worker(
                         application_error,
                         is_retryable_error);
 
-          // TODO: support multiple return values
-          // https://github.com/beacon-biosignals/Ray.jl/issues/54
-          if (return_vec.size() == 1) {
-            (*returns)[0].second = return_vec[0];
-            return Status::OK();
+          for(unsigned int i = 0; i < return_vec.size(); i++) {
+            (*returns)[i].second = return_vec[i];
           } 
-          else {
-            auto msg = "Task returned " + std::to_string(return_vec.size()) + " values. Expected 1.";
-            return Status::NotImplemented(msg);
-          };
+          return Status::OK();
         };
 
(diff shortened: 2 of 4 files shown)


src/runtime.jl
```diff
@@ -216,15 +216,15 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
...
+    oids = GC.@preserve task_args begin
         ray_jll._submit_task(fd,
                              transform_task_args(task_args),
                              serialized_runtime_env_info,
                              resources)
     end
     # CoreWorker::SubmitTask calls TaskManager::AddPendingTask which initializes
     # the local ref count to 1, so we don't need to do that here.
-    return ObjectRef(oid; add_local_ref=false)
+    return ObjectRef(only(oids); add_local_ref=false)
```
glennmoy, line 227: I'm unsure what we want to do here...

The changes so far implement "support for mulitple return variables" on the C++ side but the Julia-side remains as is and collects the `task_executor` result into a single `ObjectRef`.

i.e, in [`task_executor`](https://github.com/beacon-biosignals/Ray.jl/blob/05faae20661897adb79750400f06a6e892ec18e1/src/runtime.jl#L342C2-L343) we convert the function `result` to a `RayObject` and then `push!` this to the `returns` vector. This is completely agnostic to the type of the `result` and will always return a length-1 vector. 

The open question is what does "support multiple return variables" look like on the Julia-side?, i.e. do we want to retain current behaviour or do we want to loop over collections of `results` and `push!` them indvididually to the `returns` vector?  I'm not sure what advantages the latter gives us, and one potential downside is type instability...

kleinschmidt (changes requested):
I'm not sure I see the value of doing this (other than that we potentially _could_).  There are limited circumstances where this would actually be useful; pretty much all I can think of off the top of my head is a situation when you have a task that really has to be a _single_ task that returns a small thing and a big thing, and you want to avoid the overhead of materializing the big thing locally just to access the small thing.

The tricky thing here is that we'd  need to somehow provide the number of return values when we call `submit_task` and properly handle that through the task options creation and callback execution.  Then we have to make some tricky design decisions around what "return N values" means for us.  Do we require _exactly_ N values? (Julia destructruing doesn't, but that's a common footgun). What constitutes an "acceptable container" for these return values? (again, julia destructuring supports any iterable thing, but lots of things are iterable that you might not expect...).  Are we going to special case "single return value" (and not destructure/unpack at all) or require that these functions always return an iterable collection of some kind?

Until we have a real _need_ to support this, then I think it's better to avoid making those design decisions until necessary (and we have a bit more experience with how folks are actually using Ray in context...)

glennmoy:
> I'm not sure I see the value of doing this (other than that we potentially _could_). There are limited circumstances where this would actually be useful; pretty much all I can think of off the top of my head is a situation when you have a task that really has to be a _single_ task that returns a small thing and a big thing, and you want to avoid the overhead of materializing the big thing locally just to access the small thing.
> 
> The tricky thing here is that we'd need to somehow provide the number of return values when we call `submit_task` and properly handle that through the task options creation and callback execution. Then we have to make some tricky design decisions around what "return N values" means for us. Do we require _exactly_ N values? (Julia destructruing doesn't, but that's a common footgun). What constitutes an "acceptable container" for these return values? (again, julia destructuring supports any iterable thing, but lots of things are iterable that you might not expect...). Are we going to special case "single return value" (and not destructure/unpack at all) or require that these functions always return an iterable collection of some kind?
> 
> Until we have a real _need_ to support this, then I think it's better to avoid making those design decisions until necessary (and we have a bit more experience with how folks are actually using Ray in context...)

yep completely agree! I was initially changing this as a driveby to #190 figuring it was only tidying up something we'd left dangling since the MVP but once I hit the julia-side changes I realised there's it changes a lot more than I expected. Will close and resurrect if we ever need it again.

Issue #161, opened 2023-09-28
Automated building and publish of artifacts in CI


glennmoy:
Currently, building and publishing artifacts is a [manual process](https://github.com/beacon-biosignals/Ray.jl/blob/main/ray_julia_jll/build/README.md) that must be performed across multiple machines.

As such it is quite time-consuming and error prone, especially when `main` is updated while the PR to create artifacts is still WIP or in review. It also means that changes to `main` are not necessarily exposed to users until we've rebuilt the artifacts, for which we have no formal release process. As we move to register (a beta release of) the package, we'll be expecting to more directly support early users by adding features/bugfixes which will require more frequent releases. 

In light of the above, it would be much simpler to have these artifacts be automatically built and released via the CI such that no manual intervention is required and we can make more frequent and stable releases.

Here is a rough outline of the CI process for how this would work:
```
1. On merge to `main` a CI job will run that will build a tarball for each supported platform / Julia version (`build_tarball.jl`) then create a git tag and upload the artifacts to a pre-release (`upload_tarballs.jl`).
Each time a PR is merged the previous tag and artefacts will be overwritten.

2. The next job will run `bind_artifacts.jl`, run the tests to confirm they work, and then open a PR / commit to main (??)

3. When the Artifacts are updated in `main` the CI converts the pre-release to a release (which now can’t be overwritten).
```

The final step is to register the tag with JuliaRegistrator, which IIRC is a manual step anyway.

I'm on the fence on how we want to bind the new artifacts; opening a PR allows for verification, but requires a manual step that is perhaps unecessary? whereas pushing to `main` is quicker but potentially against our best practices.

PR #176, opened 2023-09-29
Simplify Ray.jl release process

omus:

PR #194, opened 2023-10-10
GHA workflow to automate creation of GH releases & tags

glennmoy:
Closes https://github.com/beacon-biosignals/Ray.jl/issues/161 and #207 

- deletes `upload/download_tarballs.jl` and associated code
- updates docs
- updates `CI.yml` to _not_ run on tags
- updates `artifacts_CI.yml` to _only_ run on tags (rename to Tag.yml?)
- creates `Release.yml` that is manually triggered to create (pre-)releases
- replaces `Artifacts.toml` on `main` with a dummy entry (to be used with Overrides.toml)


.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - "Project.toml"
+jobs:
+  branch:
+    name: "Create Release Branch"
```
glennmoy, line 11:
```suggestion
    name: "Publish Pre-Release"
```

.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
...
+            arch: "x64"
+        - uses: julia-actions/cache@v1
+        - name: Check Project version
+          id: version_check
+          shell: julia --color=yes --project {0}
+          run: |
+            using Pkg.Types
+            project = read_project("Project.toml")
+            latest_tag = readchomp(`git describe --tags --abbrev=0`)
+            @assert project.version == latest_tag "$(project.version) != $latest_tag"
```
glennmoy, line 29: I'm assuming this is enough to trigger a job failure?

.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
...
+            open(ENV["GITHUB_OUTPUT"], "a") do io
+              println(io, "version=$(project.version)")
+            end
+          continue-on-error: true
```
glennmoy, line 33: I want to "allow_failure" here and not have the CI fail just because the version and tag are the same

.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
...
+            using Pkg
+            Pkg.instantiate()
+            include("download_tarballs.jl")  # pulls artifacts from the latest CI job
+            include("upload_tarballs.jl")    # uploads artifacts as assets to GH prerelease
+            include("bind_artifacts.jl")     # updates Artifacts.toml with asset metadata
+        - name: Commit changes
+          run: |
+            git commit -m "Update Artifacts.toml" -- Artifacts.toml
+            git tag v${{ steps.version_check.outputs.version }}
+            git push --all origin
```
glennmoy, line 56: I assume the github credentials etc allow for this?

.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
...
+              echo "Release branch already exists on remote: $release_branch" >&2
+              exit 1
+            fi
+            git checkout -b $release_branch
+        - name: Update Artifacts.toml
+          shell: julia --color=yes --project {0}
+          run: |
+            using Pkg
+            Pkg.instantiate()
+            include("download_tarballs.jl")  # pulls artifacts from the latest CI job
```
omus, line 49: What's causing this workflow to wait for the CI workflow? At the moment these two workflows run in parallel which and this workflow would always fail. You can either do:
```yaml
on:
  workflow_run:
    workflows:
      - CI
    types: [completed]
    branches: [main]
```
And then perform the `Project.toml` filtering as part of the steps. Or you can include this as a extra job in CI and use `needs`:
```yaml
steps:
  test:
    ...
  
  release:
    needs: [test]
    ...
```
The later option also can utilize `actions/download-artifact` in the `release` job which would allow us to drop `download_tarballs.jl` entirely. An example of how that would work: https://github.com/orgs/community/discussions/26639#discussioncomment-3838014

.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
+name: Release CI
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - "Project.toml"
```
omus, line 8: When this workflow is changed on a PR you should automatically trigger as much of this workflow as possible. Only triggering this workflow on `main` means any flaws in this logic aren't caught until after the changes are merged. For this workflow we just have to avoid the `push` operation which I would make into a separate step and add a conditional `if: github.event_name != 'pull_request'`

.github/workflows/artifacts_CI.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
```
omus, line 6: Should the `Artifacts.toml` be removed from `main`?
```suggestion
```

.github/workflows/Release.yml
omus: Why not assert that `project.version >= latest_tag`?

.github/workflows/Release.yml
```diff
@@ -0,0 +1,57 @@
...
+        - name: Checkout new branch
+          if: steps.version_check.outcome = 'success'
+          run: |
+            release_branch="release-v${{ steps.version_check.outputs.version }}"
+            # Error if branch already exists
+            if [[ ! -z $(git ls-remote --heads origin $release_branch) ]]; then
+              echo "Release branch already exists on remote: $release_branch" >&2
+              exit 1
+            fi
+            git checkout -b $release_branch
```
omus, line 43: Shouldn't this step just create and push the branch to the remote and leave the checking out for another step?

.github/workflows/Release.yml
omus: Better to push only the specific things you want to push

.github/workflows/Release.yml
glennmoy: becuase we don't want the rest of the worflow to proceeed if `project.version == latest_tag`

I want this step to fail (but the pipeline to "pass") if `project.version == latest_tag`.
And I only want the rest of the workflow to run if `project.version > latest_tag`

.github/workflows/Release.yml
glennmoy: Can you elaborate? what's the difference if we checkout then push?

.github/workflows/Release.yml
glennmoy: I've restructured the workflow based on the point below of wanting it to run (as much as possible) on PRs

.github/workflows/Release.yml
glennmoy: good point - I've refactored the job pushing all "state changing" behaviour to the last step that is guarded by the `event_name` check

glennmoy:
Documenting what I've tried and learned so far:
The sticking point for now is downloading the artifacts from the workflow.
- Initially wanted to use https://github.com/actions/download-artifact but that can't download subsets of artifacts, and it automatically unpacks them all into individual folders. No option to, say, _just_ retrieve `ray_julia.*` artifacts and save them to `build/tarballs`.
- Also tried https://github.com/dawidd6/action-download-artifact which is a bit more expressive (it allows regex retrieval of artifacts) but it wasn't seeing any artifacts for the `run_id` I provided. If you leave all fields to default it fetches the latest ([but not always)](https://github.com/dawidd6/action-download-artifact/issues/240) artifacts from `main`.
- Finally, I'm using `download_tarballs.jl`. The initial problem was that the `actions/checkout` creates a new commit sha than the one associated with the job, so `download_tarballs.jl` couldn't find any job associated with the new head commit. I fixed this by modifying the action. 
- Now it seems unable to able to read the `artifacts_url` (the response comes back empty)...though it can get the `workflow_runs` response. I suspect this might be fixed by provided it with `headers`.
- Lastly, the macos CI is taking ages to run. It seems to be missing the build cache and rebuilding Ray each and every time.

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
...
+  release:
+    name: "Publish Pre-Release"
+    runs-on: ubuntu-20.04-16core
+    needs: [test]
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
omus, line 194:
```suggestion
          ref: ${{ github.event.pull_request.head.sha }} # Checkout HEAD instead of merge commit
```

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
...
+          Pkg.instantiate()
+          include(joinpath(pwd(), "bind_artifacts.jl"))
+          bind_artifacts()
+        working-directory: ${{ env.build_dir }}
+      - name: Commit Changes
+        id: commit
+        run: |
+          git checkout -b ${{ steps.version_check.outputs.release_branch }}
+          git commit -m "Update Artifacts.toml" -- Artifacts.toml
+          echo "commitish=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
```
omus, line 233: Shows this output as well as appending to the `$GITHUB_OUTPUT` file
```suggestion
          echo "commitish=$(git rev-parse HEAD)" | tee -a "$GITHUB_OUTPUT"
```

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
...
+          git checkout main
```
omus, line 234:
```suggestion
```

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
...
+      - uses: julia-actions/cache@v1
+      - name: Check Project version
+        id: version_check
+        shell: julia --color=yes --project {0}
+        run: |
+          using Pkg.Types
+          project = read_project("Project.toml")
+          open(ENV["GITHUB_OUTPUT"], "a") do io
+            println(io, "version=v$(project.version)")
+            println(io, "release_branch=release-v$(project.version)")
```
omus, line 208:
```suggestion
              println(io, "version=v$(project.version)")
              println(io, "release_branch=release-v$(project.version)")
```

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
         with:
           key: ${{ steps.build-cache.outputs.cache-primary-key }}
           path: ~/.cache
+  release:
+    name: "Publish Pre-Release"
+    runs-on: ubuntu-20.04-16core
```
omus, line 187: Don't need the big machine for this
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
...
+    needs: [test]
+    strategy:
+      fail-fast: false
```
omus, line 190: Only needed for `matrix` jobs
```suggestion
```

omus:
> Initially wanted to use https://github.com/actions/download-artifact but that can't download subsets of artifacts, and it automatically unpacks them all into individual folders. No option to, say, just retrieve `ray_julia.*` artifacts and save them to `build/tarballs`.

That is a little silly but I'm not sure why that's a deal breaker. Can't you just do `mv ray_julia.*/*.so built/tarballs` as a follow up step? Also, with what you mentioned in mind it makes more sense to combine all of the libraries into a single artifact file: https://github.com/orgs/community/discussions/26639#discussioncomment-3838014

> Also tried https://github.com/dawidd6/action-download-artifact which is a bit more expressive (it allows regex retrieval of artifacts) but it wasn't seeing any artifacts for the `run_id` I provided. If you leave all fields to default it fetches the latest (https://github.com/dawidd6/action-download-artifact/issues/240 artifacts from `main`.

I wonder if it has anything to do with this behaviour I've observed: https://github.com/beacon-biosignals/Ray.jl/pull/176#issuecomment-1741244066

> Now it seems unable to able to read the `artifacts_url` (the response comes back empty)...though it can get the `workflow_runs` response. I suspect this might be fixed by provided it with `headers`.

GitHub Actions permissions immediately comes to mind. You could try `permissions: read-all` to see if that is the problem: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes

> Lastly, the macos CI is taking ages to run. It seems to be missing the build cache and rebuilding Ray each and every time.

I've noticed this too. There is some logic which only updates the cache if the key hasn't been set. Maybe you performed a partial build on an initial build which was then cached. All future builds would be based off of that and need to do a bunch of extra work. We should definitely fix this.

omus:
> > Lastly, the macos CI is taking ages to run. It seems to be missing the build cache and rebuilding Ray each and every time.
> 
> I've noticed this too. There is some logic which only updates the cache if the key hasn't been set. Maybe you performed a partial build on an initial build which was then cached. All future builds would be based off of that and need to do a bunch of extra work. We should definitely fix this.

This does seem to be the case:

Screenshot 2023-10-12 at 12 36 54

I'll delete these caches and rebuild them by re-running the actions

.github/workflows/CI.yml
```diff
@@ -182,3 +182,66 @@ jobs:
         with:
           key: ${{ steps.build-cache.outputs.cache-primary-key }}
           path: ~/.cache
+  release:
+    name: "Publish Pre-Release"
+    runs-on: ubuntu-20.04-16core
+    needs: [test]
```
omus, line 188: As documented by https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-workflow-run-artifacts
```suggestion
    needs: [test]
    permissions:
      actions: read
```

omus:
@glennmoy I did some experimentation with `actions/download-artifacts` and ended up getting something working in https://github.com/beacon-biosignals/Ray.jl/pull/200. Feel free to review and merge that into this PR.

.github/workflows/CI.yml
glennmoy: don't think this is necessary anymore

glennmoy:
> That is a little silly but I'm not sure why that's a deal breaker. Can't you just do `mv ray_julia.*/*.so built/tarballs` as a follow up step?

Yeah when I realised it wouldn't _just work_ I started exploring the alternatives. I was making some progress with `download_tarballs.jl` but probably would have revisited if I couldn't make it work.

.github/workflows/CI.yml
```diff
@@ -182,3 +183,65 @@ jobs:
         with:
           key: ${{ steps.build-cache.outputs.cache-primary-key }}
           path: ~/.cache
```
omus, line 185: Something missing from this workflow is only triggering the `release` job when the `Project.toml` version has changed. We could put that in `release` itself but I don't want developers thinking that a release is occurring during each commit push. What I would do is either make a separate job like `release-check` or you can add this check as part of `test`. The later option just reduces noise from having extra jobs.

You would set this up with a step which checks for the `Project.toml` version change and then uses job output to see if the next job needs to run ([example](https://how.wtf/run-workflow-step-or-job-based-on-file-changes-github-actions.html#run-job-based-on-file-changes-in-github-actions)). Also, worth noting that outputs from job matrices override each other if they have the same name: https://github.com/orgs/community/discussions/26639

glennmoy:
An unfortunate side-effect of removing the Artifacts.toml from `main` is that Documenter.yml now has to build the ray_julia JLL for it to build the docs...

docs/src/building-artifacts.md
```diff
@@ -8,57 +8,12 @@ Follow [the instructions](https://github.com/beacon-biosignals/ray/blob/beacon-m
...
-   git push origin :$tag
-   git tag $tag
-
-   # Update GitHub Release to point to the updated tag
-   ```
-
-9. Register the new tag as normal with JuliaRegistrator.
+To update the Ray.jl artifacts for a new release simply open a PR which bumps the Ray.jl version in the `Project.toml` and includes any other required changes.
+The Julia CI jobs created by this PR will build the artifacts required to make a new release.
+Upon merging the PR with `main` a new tag will be created and the artifacts will be uploaded to a GitHub Pre-Release, which will be converted as a full Release once the CI for the tag passes.
```
omus, line 19:
```suggestion
To update the Ray.jl artifacts for a new release simply open a PR which bumps the Ray.jl version in the `Project.toml` and includes any other required changes. Once the PR has been merged into `main` the following actions will be run in the "CI" workflow:

1. Build the various artifacts for supported platforms.
2. Generate the `Artifacts.toml` bindings and commit them.
3. Create a new tag and GitHub pre-release which includes the artifacts as assets.

The creation of the tag will trigger the "Artifacts CI" workflow which is responsible for verifying the generated artifacts work and will promote the GitHub pre-release to a full release once the checks pass.
```

Artifacts.toml
```diff
@@ -1,38 +1,2 @@
...
-arch = "aarch64"
-git-tree-sha1 = "81cc7a633a617a01f1fc87674258d880baa4cafb"
-julia_version = "1.9.0"
-os = "macos"
-
-    [[ray_julia.download]]
-    sha256 = "5969006375abafacf2fff0a1d499b492d1a883d1529e2a20fe69d3a4908f2729"
-    url = "https://github.com/beacon-biosignals/Ray.jl/releases/download/v0.0.1/ray_julia.v0.0.1.aarch64-apple-darwin-julia_version+1.9.0.tar.gz"
+[ray_julia]
+git-tree-sha1 = "0000000000000000000000000000000000000000"
```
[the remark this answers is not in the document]
glennmoy: I thought I could butit looks for an Artifacts.toml during precompilation and crashes :(

```julia
ERROR: The following 1 direct dependency failed to precompile:

Ray [3f779ece-f0b6-4c4f-a81a-0cb2add9eb95]

Failed to precompile Ray [3f779ece-f0b6-4c4f-a81a-0cb2add9eb95] to "/Users/glenn/.julia/compiled/v1.9/Ray/jl_e6ChYh".
ERROR: LoadError: Cannot locate '(Julia)Artifacts.toml' file when attempting to use artifact 'ray_julia' in 'Ray.ray_julia_jll'
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] var"@artifact_str"(__source__::LineNumberNode, __module__::Module, name::Any, platform::Any, artifacts_toml_path::Any)
   @ Artifacts /Applications/Julia-1.9.app/Contents/Resources/julia/share/julia/stdlib/v1.9/Artifacts/src/Artifacts.jl:670
 [3] var"@artifact_str"(__source__::LineNumberNode, __module__::Module, name::Any)
   @ Artifacts /Applications/Julia-1.9.app/Contents/Resources/julia/share/julia/stdlib/v1.9/Artifacts/src/Artifacts.jl:659
 [4] include(mod::Module, _path::String)
   @ Base ./Base.jl:457
 [5] include(x::String)
   @ Ray ~/github/Ray.jl/src/Ray.jl:6
 [6] top-level scope
   @ ~/github/Ray.jl/src/Ray.jl:22
 [7] include
   @ ./Base.jl:457 [inlined]
 [8] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2049
 [9] top-level scope
   @ stdin:3
in expression starting at /Users/glenn/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:9
in expression starting at /Users/glenn/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:1
in expression starting at /Users/glenn/github/Ray.jl/src/Ray.jl:1
in expression starting at stdin:3
```

.github/workflows/Documenter.yml
```diff
@@ -28,7 +30,22 @@ jobs:
       - uses: julia-actions/setup-julia@v1
         with:
           version: "1"
+      - name: Restore build cache
+        uses: actions/cache/restore@v3
+        id: build-cache
+        with:
+          key: build-cache.ray-jl.documenter
+          path: ~/.cache
```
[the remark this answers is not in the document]
glennmoy: yeah it would be nice not to have to build the CLI...what would building a dummy library entail?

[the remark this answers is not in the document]
omus: The dummy library is probably more work than it's worth. Let's just fetch a pre-built library. It shouldn't matter which one we download

omus (approved):
@glennmoy let's update the Documenter job to no longer build the library from scratch. Beyond that this is good to merge.

.github/workflows/CI.yml
```diff
@@ -182,3 +187,84 @@ jobs:
...
+        shell: julia --color=yes --project {0}
+        env:
+          LATEST_TAG: ${{ steps.latest_tag.outputs.tag }}
+        run: |
+          using Pkg.Types
+          project = read_project("Project.toml")
+          latest_tag = parse(VersionNumber, ENV["LATEST_TAG"])
+          @info "Project version: project.version"
+          @info "Latest tag: latest_tag"
+          make_release = project.version > latest_tag
```
kleinschmidt, line 204: this is probably fine for _most_ cases, but there are times when we need to make non-sequential releases.  for instance, if there are multiple major release series that all need the same bugfix released.  i don't think we should worry about that at this point but may need to in the future.

kleinschmidt (approved):

.github/workflows/Documenter.yml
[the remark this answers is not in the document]
glennmoy: > It shouldn't matter which version we download

I'm not sure that's true, I tried pulling the latest tag and it doesn't seem precompile. I'll use the "latest artifacts" instead and see if that works

.github/workflows/CI.yml
glennmoy: very true...in that case we'll want to check that there isn't already a tag for `project.version`

.github/workflows/Documenter.yml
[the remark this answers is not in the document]
glennmoy:
```julia
$ julia --project=docs docs/make.jl                                                                                                                                  13:11:53
ERROR: LoadError: ArgumentError: invalid type for argument member in method definition for Symbol at /Users/glenn/github/Ray.jl/src/ray_julia_jll/expr.jl:37
Stacktrace:
  [1] macro expansion
    @ ~/github/Ray.jl/src/ray_julia_jll/expr.jl:37 [inlined]
  [2] top-level scope
    @ ~/github/Ray.jl/src/ray_julia_jll/common.jl:67
  [3] eval(m::Module, e::Any)
    @ Core ./boot.jl:370
  [4] top-level scope
    @ ~/github/Ray.jl/src/ray_julia_jll/common.jl:63
  [5] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
  [6] include(x::String)
    @ Ray.ray_julia_jll ~/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:1
  [7] top-level scope
    @ ~/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:16
  [8] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
  [9] include(x::String)
    @ Ray ~/github/Ray.jl/src/Ray.jl:6
 [10] top-level scope
    @ ~/github/Ray.jl/src/Ray.jl:22
 [11] include
    @ ./Base.jl:457 [inlined]
 [12] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
    @ Base ./loading.jl:2049
 [13] top-level scope
    @ stdin:3
in expression starting at /Users/glenn/github/Ray.jl/src/ray_julia_jll/common.jl:63
in expression starting at /Users/glenn/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:1
in expression starting at /Users/glenn/github/Ray.jl/src/Ray.jl:1
in expression starting at stdin:3
ERROR: LoadError: Failed to precompile Ray [3f779ece-f0b6-4c4f-a81a-0cb2add9eb95] to "/Users/glenn/.julia/compiled/v1.9/Ray/jl_8PLURY".
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, keep_loaded_modules::Bool)
   @ Base ./loading.jl:2300
 [3] compilecache
   @ ./loading.jl:2167 [inlined]
 [4] _require(pkg::Base.PkgId, env::String)
   @ Base ./loading.jl:1805
 [5] _require_prelocked(uuidkey::Base.PkgId, env::String)
   @ Base ./loading.jl:1660
 [6] macro expansion
   @ ./loading.jl:1648 [inlined]
 [7] macro expansion
   @ ./lock.jl:267 [inlined]
 [8] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1611
in expression starting at /Users/glenn/github/Ray.jl/docs/make.jl:1
```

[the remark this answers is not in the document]
glennmoy: hmm...if I turn off precomilation `julia --compiled-modules=no` it works if I run it twice from the same session

```julia
julia> include("docs/make.jl")
ERROR: LoadError: ArgumentError: invalid type for argument member in method definition for Symbol at /Users/glenn/github/Ray.jl/src/ray_julia_jll/expr.jl:37
Stacktrace:
  [1] macro expansion
    @ ~/github/Ray.jl/src/ray_julia_jll/expr.jl:37 [inlined]
  [2] top-level scope
    @ ~/github/Ray.jl/src/ray_julia_jll/common.jl:67
  [3] eval(m::Module, e::Any)
    @ Core ./boot.jl:370
  [4] top-level scope
    @ ~/github/Ray.jl/src/ray_julia_jll/common.jl:63
  [5] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
  [6] include(x::String)
    @ Ray.ray_julia_jll ~/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:1
  [7] top-level scope
    @ ~/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:16
  [8] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
  [9] include(x::String)
    @ Ray ~/github/Ray.jl/src/Ray.jl:6
 [10] top-level scope
    @ ~/github/Ray.jl/src/Ray.jl:22
 [11] include
    @ ./Base.jl:457 [inlined]
 [12] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:1840
 [13] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:1660
 [14] macro expansion
    @ ./loading.jl:1648 [inlined]
 [15] macro expansion
    @ ./lock.jl:267 [inlined]
 [16] require(into::Module, mod::Symbol)
    @ Base ./loading.jl:1611
 [17] include(fname::String)
    @ Base.MainInclude ./client.jl:478
 [18] top-level scope
    @ REPL[1]:1
in expression starting at /Users/glenn/github/Ray.jl/src/ray_julia_jll/common.jl:63
in expression starting at /Users/glenn/github/Ray.jl/src/ray_julia_jll/ray_julia_jll.jl:1
in expression starting at /Users/glenn/github/Ray.jl/src/Ray.jl:1
in expression starting at /Users/glenn/github/Ray.jl/docs/make.jl:1

julia> include("docs/make.jl")
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
[ Info: HTMLWriter: rendering HTML pages.
┌ Warning: Documenter could not auto-detect the building environment Skipping deployment.
```

glennmoy:
Spending more time than I wanted on the Documenter stuff so I'm gonna leave as is and open an issue for follow up

glennmoy:
Ok, following up after a test run the pre-release and tag were correctly made but the downstream workflow that was supposed to run on `push: tag` didn't run.

This is because GitHub precludes workflows which are triggered by a `GITHUB_TOKEN` (github actions bot) as a precaution to prevent endless workflow loops.

Looking into a few recommended solutions, one was to [create a PAT](https://github.com/orgs/community/discussions/27028), which I don't think is suitable as it's scoped to the user, while the other was to [create a  Deploy key](https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/) which is slightly better as it's scoped to the repo.

Testing the latter out didn't work either :/ so I'm going to iterate on this a little more to make a more elegant solution favouring #194 

In other words, the new workflow will be to create PRs as normal, then trigger a manual "Create Pre-Release" workflow. This circumvents the `GITHUB_TOKEN` restrictions as it's nominally triggered by the user: https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/

I've tested the Pre-release workflow on this branch `on: push` and then manually triggerd the `Artifacts CI` workflow to promote the pre-release to a Release. See below

Screenshot 2023-10-19 at 19 17 09

.github/workflows/Documenter.yml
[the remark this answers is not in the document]
glennmoy: putning on this for now https://github.com/beacon-biosignals/Ray.jl/issues/206

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+          end
+      - uses: actions/checkout@v4
+      - name: Download artifacts
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          branch: main
+          name: ray_julia_libraries
+          path: ${{ env.tarballs_dir }}
+          workflow: CI.yml
```
kleinschmidt, line 44: does this run any risk of multiple PRs being merged and picking up either too old or too new artifacts?

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Commit Changes
+        id: commit
+        run: |
+          # Alternatively environment variables can be used but both author/committer need to be set
+          # https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing
+          git config user.name beacon-buddy
+          git config user.email beacon-buddy@beacon.bio
+
+          msg="Update Artifacts.toml to use ${{ steps.project_toml.outputs.version }} assets"
+          git checkout --detach
```
kleinschmidt, line 67: I missed this first time around.  is there a reason to detach rather than branching off `main`?  I think it's better to have a direct link back to the commit that the release was cut from

omus: I added this. The logic here is that we should be using the HEAD as checked out by the previous action. What we should do is pass in the `ref` we want to use from the `workflow_dispatch` event.

omus: I'd advocate we should be using:
```yaml
- uses: actions/checkout@v3
  with:
    fetch-depth: 0  # Needed determine latest tag
    ref: main
```
If we want to be explicit (note `ref` defaults to `origin/HEAD`)

.github/workflows/Release.yml
glennmoy: actually yeah I think that's a risk if we use `branch` https://github.com/dawidd6/action-download-artifact/issues/240

but if we use `commit` it should be deterministic

omus: I believe there is a definite danger here of making a release with the wrong tarballs

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
+name: "Publish Release"
+on:
+  workflow_dispatch:
+env:
```
omus, line 4:
```suggestion
# Force a single workflow to run at a time. This ensures we don't accidentally
# try to perform concurrent publishing of releases.
concurrency: publish-release

env:
```

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+  build_dir: ./build
+  tarballs_dir: ./build/tarballs
+jobs:
+  release:
+    name: "Publish Pre-Release"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0  # Needed determine latest tag
```
omus, line 14: As you are relying on this step to persist credentials we may as well be explicit about it:
```suggestion
          fetch-depth: 0  # Needed determine latest tag
          persist-credentials: true
```

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Get Latest Tag
+        id: latest_tag
+        uses: WyriHaximus/github-action-get-previous-tag@v1
+        # TODO: in future we might need to perform backports, at which point we'll need to check
+        # that the project.version has not already been tagged
```
omus, line 20: Just curious if you tried `git fetch --tags`? If that works then you can change the check later to `!(project.version in tags)` to support backports

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+  workflow_dispatch:
+env:
+  build_dir: ./build
+  tarballs_dir: ./build/tarballs
+jobs:
+  release:
+    name: "Publish Pre-Release"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
omus, line 12:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+          project = read_project("Project.toml")
+          latest_tag = parse(VersionNumber, ENV["LATEST_TAG"])
+          @info "Project version: project.version"
+          @info "Latest tag: latest_tag"
+          project.version > latest_tag || exit(1)  # error if can't make a tag
+          open(ENV["GITHUB_OUTPUT"], "a") do io
+              println(io, "version=$(project.version)")
+              println(io, "tag=v$(project.version)")
+          end
+      - uses: actions/checkout@v4
```
omus, line 37:
```suggestion
```

.github/workflows/Release.yml
glennmoy: added

.github/workflows/Release.yml
```diff
@@ -0,0 +1,82 @@
...
+        with:
+          commit: ${{ github.sha }}  # pull the artifacts for the last commit on main
+          name: ray_julia_libraries
+          path: ${{ env.tarballs_dir }}
+          workflow: CI.yml
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: "1.8"
+          arch: "x64"
+      - uses: julia-actions/cache@v1
```
omus, line 50: You don't want to share the same Julia cache as the CI actions:
```suggestion
      - uses: julia-actions/cache@v1
        with:
          cache-name: "${{ github.workflow }}-${{ github.job }}"
```

.github/workflows/Release.yml
glennmoy: I considered having an `input` for the `workflow_dispatch` but I couldn't see in what scenario we wouldn't want to tag off `main`...unless we want to tag from a specific commit on `main` in case there's been another merge in the meantime?

.github/workflows/Release.yml
```diff
@@ -0,0 +1,83 @@
+name: "Publish Release"
```
omus, line 1: Also rename this file to `Pre-release.yml`
```suggestion
name: "Publish Pre-release"
```

.github/workflows/Release.yml
omus: > I considered having an `input` for the `workflow_dispatch` but I couldn't see in what scenario we wouldn't want to tag off `main`...unless we want to tag from a specific commit on `main` in case there's been another merge in the meantime?

The scenario would be backports. I think we can punt on this for now.

.github/workflows/Release.yml
glennmoy: yeah I think it worked - I'll do it as a follow up. Afraid of accidentally breaking things now...

omus (approved):
Left another round of suggestions. I'd prefer if the Pre-release workflow was automatically triggered but this will do. 

I believe the issue you ran into with the push not re-triggering the workflow is this: https://github.community/t/push-from-action-does-not-trigger-subsequent-action/16854. I don't think we need to take care of that in this PR and can look into this as a follow up.

.github/workflows/Release.yml
glennmoy: 💡

PR #197, opened 2023-10-11
Use Julia depot stacking in our Dockerfile

omus:
Refactoring this Dockerfile again to make it friendly for stacking Julia depots. Mainly this is useful for other Dockerfiles which want to copy out the Ray.jl installation for use in another image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -188,7 +188,7 @@ RUN --mount=type=cache,target=${BAZEL_CACHE},sharing=locked,uid=${UID},gid=${GID
 
 # Copy over artifacts generated during the previous stages
 COPY --chown=${UID} --from=deps --link ${JULIA_USER_DEPOT} ${JULIA_USER_DEPOT}
-RUN rm -f ~/.julia && \
+RUN rm -rf ~/.julia && \
     ln -sf ${JULIA_USER_DEPOT} ~/.julia
 
 # Setup ray_julia library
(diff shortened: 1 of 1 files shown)


omus:
I've validated that these changes work on the private multi-node demo I have setup. Issues with CI are CI specific so I'll work through those here.

Dockerfile
```diff
@@ -82,21 +94,23 @@ RUN --mount=type=cache,sharing=locked,target=${JULIA_DEPOT_CACHE},uid=${UID},gid
...
 #####
 
-FROM ray-base as ray-jl
+FROM ray-base as build-ray-jl
 
 # Install Bazel and compilers
 ARG BAZEL_CACHE=/mnt/bazel-cache
-RUN set -eux; \
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
```
glennmoy, line 112: why the additiona of new cache targets?

glennmoy (commented):
gonna have to re-review this to wrap my head around it

Dockerfile
```diff
@@ -208,14 +224,25 @@ COPY --chown=${UID} . ${RAY_JL_PROJECT}/
 
 # Restore content from previous build directory
 RUN rm -rf ${BUILD_PROJECT} && \
-    ln -s ${BUILD_ROOT} ${BUILD_PROJECT}
+    mv ${BUILD_ROOT} ${BUILD_PROJECT}
```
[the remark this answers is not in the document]
omus: Good question. We now move the `BUILD_ROOT` into the `BUILD_PROJECT` as this will include the content in the `JULIA_USER_DEPOT`. Using a symlink would mean when we do `COPY --from=build-ray-jl ${JULIA_USER_DEPOT} ${JULIA_USER_DEPOT}` the `BUILD_ROOT` would not also be copied.

Dockerfile
omus: This is just an optimization and slightly improves performance when running these steps. Based upon https://docs.docker.com/engine/reference/builder/#example-cache-apt-packages. Looks like I missed a setting as part of this which I'll add in now

glennmoy (approved):

PR #198, opened 2023-10-12
Always update CI build cache

omus:
In response to: https://github.com/beacon-biosignals/Ray.jl/pull/194#issuecomment-1760074998

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -178,7 +178,7 @@ jobs:
       # https://github.com/actions/cache/tree/main/save#always-save-cache
       - name: Save build cache
         uses: actions/cache/save@v3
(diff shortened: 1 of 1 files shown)


omus:
Need to think about this more. Possibly we only want to save the cache when the build steps succeed. It could be helpful to save the cache on partial builds though. Mainly what we want to avoid is not-saving the cache when we save a bad cache state.

omus:
Some ideas:

- https://docs.github.com/en/actions/learn-github-actions/expressions#always
- https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

One idea I had is caching pass/failures separately so that we know that if we had a hit on a failure cache we should create a pass cache. The logic for this is a bit more complicated unfortunately

omus:
Closing this for now as it's not a priority at this moment. Still annoying but there are other priorities.

PR #201, opened 2023-10-13
Handle null-terminated `ObjectRef` fields

omus:
Potential fix to https://github.com/beacon-biosignals/Ray.jl/issues/177. It appears that the registered address of some `ObjectRef` is [being corrupted](https://github.com/beacon-biosignals/Ray.jl/issues/177#issuecomment-1758311004) most likely through bad CxxWrap handling of null-terminated strings.

I took reviewed our `ObjectRef` code and found a few places where we convert between `StdString` and `String` where a null character could result in an invalid string being produced.

I'll call out I don't know for certain that this will address the problem but it definitely won't hurt.


omus:
Closing in favor of #203

Issue #52, opened 2023-08-29
Use RayConfig (or similar) to get runtime info instead of parsing raylet logs


glennmoy:

glennmoy:
Looking through the python code the [`ray.init()` function](https://github.com/ray-project/ray/blob/56affb7e4b5af8b1da7a322fdc6ecc7e2258d522/python/ray/_private/worker.py#L1136) contains a mix of behaviours depending on whether it's connecting to an existing cluster or not... we want to look at the case where it's [connecting to an existing cluster](https://github.com/ray-project/ray/blob/56affb7e4b5af8b1da7a322fdc6ecc7e2258d522/python/ray/_private/worker.py#L1571)...

Note: it may be sufficient to simply search for the address of an existing cluster.  The cluster `address` is easy enough to get if not provided, it lives in a new file:
> If no address is provided, try to find an existing Ray instance to connect to. This is done by first checking the environment variable `RAY_ADDRESS`. If this is not defined, check the address of the latest cluster started (found in /tmp/ray/ray_current_cluster) if available. If this is also empty, then start a new local Ray instance.

This branch is triggered if [`bootstrap_address`](https://github.com/ray-project/ray/blob/56affb7e4b5af8b1da7a322fdc6ecc7e2258d522/python/ray/_private/worker.py#L1481C1-L1486C89) is parsed from [`canonicalize_bootstrap_address`](https://github.com/ray-project/ray/blob/56affb7e4b5af8b1da7a322fdc6ecc7e2258d522/python/ray/_private/services.py#L516-L538)

```python
redis_address, gcs_address = None, None
    bootstrap_address = services.canonicalize_bootstrap_address(address, _temp_dir)
    if bootstrap_address is not None:
        gcs_address = bootstrap_address
        logger.info("Connecting to existing Ray cluster at address: %s...", gcs_address)
```

which internally calls [`get_ray_address_from_environment`](https://github.com/ray-project/ray/blob/56affb7e4b5af8b1da7a322fdc6ecc7e2258d522/python/ray/_private/services.py#L383-L429). This does a few things but the ones that interest us are:
[...]

PR #202, opened 2023-10-16
Use ray-project fork in CI build

glennmoy:
Uses our [upstream branch of the ray project](https://github.com/ray-project/ray/pull/40098) instead of our forked repo in the CI and docker file


glennmoy:
A bit of wrinkle in getting this to work is that the ray args can longer be parsed from `raylet.out` as the full commandisn't reported there anymore. We might be able to still parse some args from `raylet.out` or elsewhere but it might also be better to revisit #52

glennmoy:
Current issue is that newer versions of RAy ship with the autogenereated protobuf files checked in as of https://github.com/ray-project/ray/pull/39210, and rebuilding ray from source _doesn't_ seem to regenerate them.

Thus it isn't picking up on the `Language::JULIA` support that we added to the source files.
As a resuilt it cannot spawn worker processes and the `task.jl` tests fail.

glennmoy:
Rebased against https://github.com/ray-project/ray/commit/68d1feb871d5136a923141ff9e57c751add22eff

PR #203, opened 2023-10-17
Improve robustness of `ObjectRef` ownership info serialization

omus:
While investigating #177 it was becoming evident that the flaw had something do with our worker addresses once again. As I so far have been unable to reproduce this problem locally I decided to look into audit our object reference ownership code to see if anything popped out. In doing so I recalled that in #140 we utilized JSON serialization for the owner address as we were having problems with CxxWrap `StdString` which would interpret the null-character as the end of the string instead of looking at the length of the string. The JSON approach doesn't actually fix anything except that it works around the issue. I decided to try to switch back to using `SerializeAsString` and apply our new CxxWrap knowledge to see if we could really fix the issue. In doing so I discovered that CxxWrap doesn't handle non-ASCII characters well:

```julia
julia> str = "¿\0ÿ"
"¿\0ÿ"

julia> collect(StdString(str, ncodeunits(str)))
5-element Vector{Char}:
 'Â': Unicode U+00C2 (category Lu: Letter, uppercase)
 '¿': Unicode U+00BF (category Po: Punctuation, other)
 '\0': ASCII/Unicode U+0000 (category Cc: Other, control)
 'Ã': Unicode U+00C3 (category Lu: Letter, uppercase)
 '¿': Unicode U+00BF (category Po: Punctuation, other)

julia> std_str = StdString(str, ncodeunits(str))
"¿"

julia> collect(String(std_str))
1-element Vector{Char}:
 '¿': Unicode U+00BF (category Po: Punctuation, other)

julia> collect(String(collect(std_str)))
5-element Vector{Char}:
 'Â': Unicode U+00C2 (category Lu: Letter, uppercase)
 '¿': Unicode U+00BF (category Po: Punctuation, other)
 '\0': ASCII/Unicode U+0000 (category Cc: Other, control)
 'Ã': Unicode U+00C3 (category Lu: Letter, uppercase)
 '¿': Unicode U+00BF (category Po: Punctuation, other)

julia> collect(String(Vector{UInt8}(collect(std_str))))
3-element Vector{Char}:
 '¿': Unicode U+00BF (category Po: Punctuation, other)
 '\0': ASCII/Unicode U+0000 (category Cc: Other, control)
 'ÿ': Unicode U+00FF (category Ll: Letter, lowercase)
```

[...]

PR #204, opened 2023-10-17
Store ownership info in `RaySerializer`

omus:
Follow up to #203. Storing the owner address and the serialized object status in `ObjectRef` was rather strange as this information was only needed to finish the registration process after deserialization. This PR changes things to no longer be stored in `ObjectRef` but as part of the `RaySerializer` in a lookup table of object ref to ownership data.

Mainly this is just a refactoring of the existing code which should make it clearer that the extra `ObjectRef` fields are not needed all the time.

### DIFF
--- a/src/ray_serializer.jl
+++ b/src/ray_serializer.jl
@@ -1,3 +1,8 @@
+struct OwnershipInfo
+    owner_address::ray_jll.Address
+    serialized_object_status::String
+end
+
 mutable struct RaySerializer{I<:IO} <: AbstractSerializer
     # Fields required by all AbstractSerializers
     io::I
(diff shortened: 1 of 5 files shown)


src/ray_serializer.jl
```diff
@@ -110,7 +139,9 @@ function deserialize_from_ray_object(ray_obj::SharedPtr{ray_jll.RayObject},
     end
 
     for inner_object_ref in s.object_refs
-        _register_ownership(inner_object_ref, outer_object_ref)
+        metadata = s.object_owner[inner_object_ref]
+        _register_ownership(inner_object_ref, outer_object_ref, metadata.owner_address,
+                            metadata.serialized_object_status)
```
kleinschmidt, line 144: could even do property name destructuring here if you wanted (I dunno if it makes things any clearer or not though)

kleinschmidt (approved):
nice!

src/ray_serializer.jl
omus: I'll take a quick look to see if that makes things more readable

omus: Seems about the same. I'll update this to use the destructuring for no real reason beyond I want to

kleinschmidt: 😎

Issue #55, opened 2023-08-29
Do not use sizeof in construcing LocalMemoryBuffer


glennmoy:
Using `sizeof` is probably wrong for `pointer_from_objref` as there is probably
additional Julia type metadata not being accounted for. This may be why we see a
segfault when trying to use `pointer_from_objref`.

omus:
This issus is referring to this testset: https://github.com/beacon-biosignals/Ray.jl/blob/fc586a252ff4d0312b5731ce3cd43c48db8c94ab/test/ray_julia_jll/buffer.jl#L21-L35

The comment isn't actually something to fix but is just a note about why copying data from a `pointer_from_objref` doesn't work. I'll revise the comments to be clearer and close this.

PR #205, opened 2023-10-18
Comment on use of `pointer_from_objref` in `LocalMemoryBuffer` tests

omus:
Closes: https://github.com/beacon-biosignals/Ray.jl/issues/55

### DIFF
--- a/test/ray_julia_jll/buffer.jl
+++ b/test/ray_julia_jll/buffer.jl
@@ -5,6 +5,10 @@ using .ray_julia_jll: LocalMemoryBuffer, Data, Size, OwnsData, IsPlasmaBuffer
 @testset "LocalMemoryBuffer" begin
     data = UInt16[1:3;]
 
+    # In practise no one should be using `pointer_from_objref` with Ray buffers. We only do
+    # so as part of these tests to demonstrate that basic buffer functionality works without
+    # having to utilize serialization.
+
     @testset "non-copied object reference" begin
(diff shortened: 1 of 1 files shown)


test/ray_julia_jll/buffer.jl
```diff
@@ -18,11 +22,10 @@ using .ray_julia_jll: LocalMemoryBuffer, Data, Size, OwnsData, IsPlasmaBuffer
...
     end
 
-    # TODO: Using `sizeof` is probably wrong for `pointer_from_objref` as there is probably
-    # additional Julia type metadata not being accounted for. This may be why we see a
-    # segfault when trying to use `pointer_from_objref`.
-    # https://github.com/beacon-biosignals/Ray.jl/issues/55
     @testset "copied object reference" begin
+        # Using `sizeof` is probably wrong for `pointer_from_objref` as there is probably
+        # additional Julia type metadata not being accounted for. This may be why we see a
+        # segfault when trying to use `pointer_from_objref`.
```
kleinschmidt, line 28:
```suggestion
        # Using `sizeof` is probably wrong for `pointer_from_objref` as there is probably
        # additional Julia type metadata not being accounted for. This may be why we see a
        # segfault when trying to use `pointer_from_objref`.  However as noted above, no
        # one should be using this in practice, and it's only used here to test buffer
        # functionality separately from testing serialization.
```

kleinschmidt (commented):
just to be suuuuper clear

PR #211, opened 2023-10-20
Use GCSClient instead of PythonGCSClient

### LINKED ISSUE #76: Stop using the `PythonGcsClient`
This GCS client specifically mentions it should only be used by Python:

https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/gcs/gcs_client/gcs_client.h#L187-L188

glennmoy:
Break off of https://github.com/beacon-biosignals/Ray.jl/pull/202

Closes #76 

Two things worth calling out:
- Unlike the `PythonGcsClient` the `GcsClient` has no direct interface for specifying a `timeout`. This is instead set through `RayConfig` parameters which can be overriden via environment variables.
- Attempting to `Connect` to unintiated GCS Servers issues a warning message before killing the driver process after a 1-min timeout.


test/function_manager.jl
```diff
@@ -101,4 +104,5 @@
     # finally
     #     rmprocs(workers())
     # end
+    Disconnect(client)
```
glennmoy, line 107: probably better to us `finally` or somethign?

build/wrapper.cc
```diff
@@ -257,62 +257,68 @@ JuliaGcsClient::JuliaGcsClient(const std::string &gcs_address) {
...
     std::string value;
-    Status status = gcs_client_->InternalKVGet(ns, key, timeout_ms, value);
+    Status status = gcs_client_->InternalKV().Get(ns, key, value);
     if (!status.ok()) {
         throw std::runtime_error(status.ToString());
     }
     return value;
 }
 
-int JuliaGcsClient::Put(const std::string &ns,
```
glennmoy, line 278: I kept the original signature for these rather than return `Status` as it made for less changes to the code

test/ray_julia_jll/gcs_client.jl
```diff
@@ -1,44 +1,39 @@
...
     # throw on missing key
-    @test_throws ErrorException Get(client, ns, "none", -1)
+    @test_throws ErrorException Get(client, ns, "none")
 
-    # ideally we'd throw on connect but it returns OK......
-    badclient = JuliaGcsClient("127.0.0.1:6378")
-    status = Connect(badclient)
-
-    # ...but then throws when we try to do anything so at least there's that
-    @test_throws ErrorException Put(badclient, ns, "computer", "mistaek", false, -1)
```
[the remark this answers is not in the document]
glennmoy: MWE
```julia
julia> using Ray.ray_julia_jll

julia> badclient = ray_julia_jll.JuliaGcsClient("127.0.0.1:6378")
Ray.ray_julia_jll.JuliaGcsClientAllocated(Ptr{Nothing} @0x00000001376270e0)

julia> ray_julia_jll.Connect(badclient)
[2023-10-20 18:04:32,822 E 28309 8652849] gcs_rpc_client.h:207: Failed to connect to GCS at address 127.0.0.1:6378 within 5 seconds.
OK

julia> [2023-10-20 18:05:27,877 E 28309 8653025] gcs_rpc_client.h:537: Failed to connect to GCS within 60 seconds. GCS may have been killed. It's either GCS is terminated by `ray stop` or is killed unexpectedly. If it is killed unexpectedly, see the log file gcs_server.out. https://docs.ray.io/en/master/ray-observability/ray-logging.html#logging-directory-structure. The program will terminate.
(venv) MacBook-Air~/.j/d/Ray (gm/gcsclient|✔)

File: build/wrapper.cc [hunk 4]
```diff
@@ -257,62 +257,68 @@ JuliaGcsClient::JuliaGcsClient(const std::string &gcs_address) {
...
-std::vector<std::string> JuliaGcsClient::Keys(const std::string &ns,
-                                              const std::string &prefix,
-                                              int64_t timeout_ms) {
+std::vector<std::string> JuliaGcsClient::Keys(const std::string &ns, const std::string &prefix) {
     if (!gcs_client_) {
         throw std::runtime_error("GCS client not initialized; did you forget to Connect?");
     }
     std::vector<std::string> results;
-    Status status = gcs_client_->InternalKVKeys(ns, prefix, timeout_ms, results);
+    Status status = gcs_client_->InternalKV().Keys(ns, prefix, results);
```
omus, line 309: Looks like `gcs_client_->InternalKV().AsyncInternalKVKeys(...)` may be the thread to pull on for specifying a non-ENV timeout. I noticed `AsyncGetByName` in the `accessor.h` file which does have a timeout. It's unclear why some `Async...` methods have timeouts and other's don't

src/function_manager.jl
```diff
@@ -65,8 +65,10 @@ const FUNCTION_MANAGER = Ref{FunctionManager}()
 function _init_global_function_manager(gcs_address)
     @info "Connecting function manager to GCS at $gcs_address..."
     gcs_client = ray_jll.JuliaGcsClient(gcs_address)
-    ray_jll.Connect(gcs_client)
+    status = ray_jll.Connect(gcs_client)
+    ray_jll.ok(status) || error("Could not connect to GCS")
     FUNCTION_MANAGER[] = FunctionManager(; gcs_client, functions=Dict{String,Any}())
+    atexit(() -> ray_jll.Disconnect(gcs_client))
```
omus, line 71: Probably makes more sense to use a `finalizer` on `FunctionManager`

src/function_manager.jl
```diff
@@ -82,32 +84,16 @@ function export_function!(fm::FunctionManager, f, job_id=get_job_id())
...
-    catch e
-        if e isa ErrorException && contains(e.msg, "Deadline Exceeded")
-            return :timed_out
-        else
-            rethrow()
-        end
-    end
-    return status
-end
-
```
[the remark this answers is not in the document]
omus: > none of the tests actually cared about the timeout I'm wondering if it's worth the hassle?

I agree with your assessment. Would be good to add some comments about this timeout information as well as any scenarios you've noticed where a timeout could be helpful.

test/function_manager.jl
omus: Using `finally` would be better. Alternatively you could use a `do`-block like this:
```julia
Connect(JuliaGcsClient("127.0.0.1:6379")) do client
   ...
end
```

test/ray_julia_jll/gcs_client.jl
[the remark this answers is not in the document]
omus: The scenario in which this would occur is if you run `Ray.init()` without a local raylet present? If so, I'm fine with making this into an issue to tackle later.

test/ray_julia_jll/gcs_client.jl
```diff
@@ -1,44 +1,39 @@
...
+    Disconnect(client)
```
omus, line 38: Another case we should always disconnect

omus (commented):
Overall looks good. The C++ code was a bit hard to follow having not dived into the GCS myself. I think adding more comments and links in the C++ code would be helpful. I agree supporting timeouts isn't worth the effort at this time as we haven't been using it anyway.

test/ray_julia_jll/gcs_client.jl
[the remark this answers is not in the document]
glennmoy: yeah and even then it should error earlier if it tried to get the GCS address during `Ray.init`

build/wrapper.cc
glennmoy: I think it's limited to Actors stuff? Only `AsyncRegisterActor`, `AsyncListNamedActors`, `AsyncGetByName`  have `timeout_ms` as a paramter.

`InternalKVKeys` doesn't: https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/gcs/gcs_client/accessor.h#L698C1-L698C1

omus: The `timeout_ms` is limited to actors but if you did really want it I was saying these `Async` methods would allow you to specify it. For example: https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/src/ray/gcs/gcs_client/accessor.cc#L1092-L1110

glennmoy: sorry I understand now! I think it's ok to punt for now? we know where to look at least if we want it later?

src/function_manager.jl
glennmoy: added

build/wrapper.cc
omus: You can definitely punt on it for now. I just wanted to mention this option.

test/function_manager.jl
glennmoy: added a new method with do-syntax

test/ray_julia_jll/gcs_client.jl
glennmoy: I wrapped this up in the `Connect(f, client)` method.

src/function_manager.jl
```diff
@@ -55,19 +55,26 @@ end
...
 const FUNCTION_MANAGER_NAMESPACE = "jlfun"
 
-Base.@kwdef struct FunctionManager
+Base.@kwdef mutable struct FunctionManager
     gcs_client::ray_jll.JuliaGcsClient
     functions::Dict{String,Any}
+
+    function FunctionManager(gcs_client, functions)
+        status = ray_jll.Connect(gcs_client)
+        ray_jll.ok(status) || error("Could not connect to GCS")
```
glennmoy, line 64: a bit opinionated maybe? I figured with the `finalizer` calling `Disconnect` we were implicitly tying the `Connect` to the constructor?

test/ray_julia_jll/gcs_client.jl
```diff
@@ -1,44 +1,35 @@
 @testset "GCS client" begin
     using UUIDs
     using .ray_julia_jll: JuliaGcsClient, Connect, Put, Get, Keys, Exists, Status, ok,
-                          ToString
+                          ToString, Disconnect
 
     client = JuliaGcsClient("127.0.0.1:6379")
```
omus, line 6: Add a test for calling `Disconnect` on non-connected clients

src/ray_julia_jll/common.jl
```diff
@@ -79,6 +79,17 @@ function check_status(status::Status)
...
+function Connect(body, client::JuliaGcsClient)
+    status = Connect(client)
+    ok(status) || error("Could not connect to GCS")
+    try
+        body(client)
+    finally
+        Disconnect(client)
+    end
+    return nothing
+end
```
omus, line 91: Might as well return the last statement from `body`:
```suggestion
    try
        return body(client)
    finally
        Disconnect(client)
    end
end
```

src/ray_julia_jll/common.jl
```diff
@@ -79,6 +79,17 @@ function check_status(status::Status)
     return nothing
 end
 
+function Connect(body, client::JuliaGcsClient)
+    status = Connect(client)
+    ok(status) || error("Could not connect to GCS")
```
omus, line 84: Show the status in the error?

src/function_manager.jl
```diff
@@ -55,19 +55,25 @@ end
...
+    end
 end
 
 const FUNCTION_MANAGER = Ref{FunctionManager}()
 
 function _init_global_function_manager(gcs_address)
     @info "Connecting function manager to GCS at $gcs_address..."
     gcs_client = ray_jll.JuliaGcsClient(gcs_address)
-    ray_jll.Connect(gcs_client)
     FUNCTION_MANAGER[] = FunctionManager(; gcs_client, functions=Dict{String,Any}())
```
omus, line 76:
```suggestion
    ray_jll.Connect(gcs_client)
    finalizer(ray_jll.Disconnect, gcs_client)
    FUNCTION_MANAGER[] = FunctionManager(; gcs_client, functions=Dict{String,Any}())
```

src/function_manager.jl
omus: I think this is a step too far as now it's impossible to create a `FunctionManager` without connecting it to the GCS client. I can understand your concern with running `Disconnect` automatically on every `FunctionManager`. What if you just use `finalizer` on a single instance created in `_init_global_function_manager`?

omus (commented):
Only the `FunctionManager` changes are blocking approval

src/function_manager.jl
glennmoy: ah, so I didn't realise that was possible. the docs kinda suggest that it should be called on instantiatoin but perhaps that was my misunderstanding

glennmoy: I'll rever this change and add the `finalizer` as you suggest

src/ray_julia_jll/common.jl
glennmoy: I've called `check_status` instead

test/ray_julia_jll/gcs_client.jl
glennmoy: huh, segfault.. I'll look into this

glennmoy: needed to protect Disconnecting in the wrapper code

src/function_manager.jl
```diff
@@ -55,19 +55,26 @@ end
...
 
 const FUNCTION_MANAGER = Ref{FunctionManager}()
 
 function _init_global_function_manager(gcs_address)
     @info "Connecting function manager to GCS at $gcs_address..."
     gcs_client = ray_jll.JuliaGcsClient(gcs_address)
-    ray_jll.Connect(gcs_client)
+    status = ray_jll.Connect(gcs_client)
+    ray_jll.check_status(status)
+    finalizer(ray_jll.Disconnect, gcs_client)
```
omus, line 76: Adding the `finalizer` earlier in case `the `ray_jll.check_status` ends the function early
```suggestion
    finalizer(ray_jll.Disconnect, gcs_client)
    status = ray_jll.Connect(gcs_client)
    ray_jll.check_status(status)
```

src/function_manager.jl
```diff
@@ -55,19 +55,26 @@ end
...
+Base.@kwdef mutable struct FunctionManager
     gcs_client::ray_jll.JuliaGcsClient
     functions::Dict{String,Any}
+
+    function FunctionManager(gcs_client, functions)
+        fm = new(gcs_client, functions)
+        f(x) = ray_jll.Disconnect(x.gcs_client)
+        return finalizer(f, fm)
+    end
 end
```
omus, line 67: I don't think we need the `finalizer` at all here now.
```suggestion
Base.@kwdef struct FunctionManager
    gcs_client::ray_jll.JuliaGcsClient
    functions::Dict{String,Any}
end
```

omus (approved):

src/function_manager.jl
glennmoy: ah yeah forgot to remove it

glennmoy: pretty sure the `FunctionManager` needs to be mutable

omus:
Failures definitely look related:
```
 [ Info: Connecting function manager to GCS at 10.0.0.105:6379...
terminate called after throwing an instance of 'std::runtime_error'
  what():  NotFound: Failed to find the key

signal (6): Aborted
in expression starting at /home/runner/work/Ray.jl/Ray.jl/test/ray_serializer.jl:1
pthread_kill at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
raise at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
```
– https://github.com/beacon-biosignals/Ray.jl/actions/runs/6645852901/job/18058052127?pr=211#step:12:215

```
 [ Info: Connecting function manager to GCS at 127.0.0.1:6379...
libc++abi: terminating due to uncaught exception of type std::runtime_error: NotFound: Failed to find the key

signal (6): Abort trap: 6
in expression starting at /Users/runner/work/Ray.jl/Ray.jl/test/ray_serializer.jl:81
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 44751822 (Pool: 44725417; Big: 26405); GC: 51
```
– https://github.com/beacon-biosignals/Ray.jl/actions/runs/6645852901/job/18058052370?pr=211#step:12:211

omus:
I can manage to reproduce this on my local machine:
```
┌ Debug: Skipping registering ownership for ObjectRef("00ffffffffffffffffffffffffffffffffffffff4300000013000000") as object has known owner
└ @ Ray ~/.julia/dev/Ray/src/object_ref.jl:128
libc++abi: terminating due to uncaught exception of type std::runtime_error: NotFound: Failed to find the key

[1307] signal (6): Abort trap: 6
in expression starting at /Users/cvogt/.julia/dev/Ray/test/runtests.jl:19
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 30846806 (Pool: 30821062; Big: 25744); GC: 44
ERROR: Package Ray errored during testing (received signal: 6)
```

omus:
The issue noticed above was just due to us failing to `Disconnect` the `JuliaGcsClient` used in the `function_manager.jl` tests. I've added a destructor to the `JuliaGcsClient` to automatically disconnect the client to avoid the SIGABRT failure. However, since users of `JuliaGcsClient` should really be handling disconnecting themselves I added in a STDERR warning from C++ indicating that a connected `JuliaGcsClient` failed to disconnect on their own. As Ray.jl users aren't exposed to this class mostly this is just a reminder for Ray.jl developers to be explicit about disconnecting.

glennmoy:
> The issue noticed above was just due to us failing to `Disconnect` the `JuliaGcsClient` used in the `function_manager.jl` tests. I've added a destructor to the `JuliaGcsClient` to automatically disconnect the client to avoid the SIGABRT failure. However, since users of `JuliaGcsClient` should really be handling disconnecting themselves I added in a STDERR warning from C++ indicating that a connected `JuliaGcsClient` failed to disconnect on their own. As Ray.jl users aren't exposed to this class mostly this is just a reminder for Ray.jl developers to be explicit about disconnecting.

yeah when I removed the `finalizer` from the inner constructor as per your suggestion above I should have done so by reverting https://github.com/beacon-biosignals/Ray.jl/pull/211/commits/6b98717caa778604f27918461dc30f6cc7b3d9bf as I missed adding back the `Disconnect`

PR #214, opened 2023-10-20
Parse ray runtime parameters from GlobalStateAccessor

### LINKED ISSUE #52: Use RayConfig (or similar) to get runtime info instead of parsing raylet logs

glennmoy:
Break off of #202 

Closes #52 

- gcs address: `/tmp/ray/ray_current_cluster`
- node ip address: [const "127.0.0.1](https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/python/ray/_private/services.py#L650-L658) (later it comes from its own file)
- raylet, store, port: GlobalStateAccessor

---
  - https://app.asana.com/0/0/1205744051874128
  - https://app.asana.com/0/0/1205770557599276

### DIFF
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -46,3 +46,5 @@ const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
 # In ray-2.5.1 this is constant but in later versions it's read from NODE_IP_ADDRESS.json
 # https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/python/ray/_private/services.py#L650-L658
 const NODE_IP_ADDRESS = "127.0.0.1"
+
+const DEFAULT_SESSION_DIR = "/tmp/ray/session_latest"
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -1,0 +1,48 @@
+# NOTES:
+#
+# python function manager maintains a local table of "execution info" with a
+# "function_id" key and values that are named tuples of name/function/max_calls.
+#
+# python remote function sets a UUID4 at construction time:
+# https://github.com/beacon-biosignals/ray/blob/7ad1f47a9c849abf00ca3e8afc7c3c6ee54cda43/python/ray/remote_function.py#L128
+#
+# ...that's used to set the function_hash (???)...
+# https://github.com/beacon-biosignals/ray/blob/7ad1f47a9c849abf00ca3e8afc7c3c6ee54cda43/python/ray/remote_function.py#L263-L265
+#
+# later comment suggests that "ideally" they'd use the hash of the pickled
+# function:
+# https://github.com/beacon-biosignals/ray/blob/7ad1f47a9c849abf00ca3e8afc7c3c6ee54cda43/python/ray/includes/function_descriptor.pxi#L183-L186
+#
+# ...but that it's not stable for some reason.  but.....neither is a random
+# UUID?????
+#
+# the function table key is built like
+# <key type>:<jobid>:key
+
+# function manager holds:
+# local cache of functions (keyed by function id/hash from descriptor)
+# gcs client
+# ~~maybe job id?~~ this is managed by the core worker process
+
+# https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/ray_constants.py#L122-L123
+const FUNCTION_SIZE_WARN_THRESHOLD = 10_000_000  # in bytes
+const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000  # in bytes
+
+# python uses "fun" for the namespace: https://github.com/beacon-biosignals/ray/blob/7ad1f47a9c849abf00ca3e8afc7c3c6ee54cda43/python/ray/_private/ray_constants.py#L380
+# so "jlfun" seems reasonable
+const FUNCTION_MANAGER_NAMESPACE = "jlfun"
+
+# https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/utils.py#L744-L746
+const _check_msg = "Check that its definition is not implicitly capturing a large " *
+                   "array or other object in scope. Tip: use `Ray.put()` to put large " *
+                   "objects in the Ray object store."
+
+
+# env var to control whether logs are sent do stderr or to file.  if "1", sent
+# to stderr; otherwise, will be sent to files in `/tmp/ray/session_latest/logs/`
+# https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/ray_constants.py#L198
+const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
+
+# In ray-2.5.1 this is constant but in later versions it's read from NODE_IP_ADDRESS.json
+# https://github.com/ray-project/ray/blob/a03efd9931128d387649dd48b0e4864b43d3bfb4/python/ray/_private/services.py#L650-L658
+const NODE_IP_ADDRESS = "127.0.0.1"
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -32,11 +32,6 @@ const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000  # in bytes
 # so "jlfun" seems reasonable
 const FUNCTION_MANAGER_NAMESPACE = "jlfun"
 
-# https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/utils.py#L744-L746
-const _check_msg = "Check that its definition is not implicitly capturing a large " *
-                   "array or other object in scope. Tip: use `Ray.put()` to put large " *
-                   "objects in the Ray object store."
-
 # env var to control whether logs are sent do stderr or to file.  if "1", sent
 # to stderr; otherwise, will be sent to files in `/tmp/ray/session_latest/logs/`
 # https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/ray_constants.py#L198
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -48,3 +48,5 @@ const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
 const NODE_IP_ADDRESS = "127.0.0.1"
 
 const DEFAULT_SESSION_DIR = "/tmp/ray/session_latest"
+
+const GCS_ADDRESS_FILE = "/tmp/ray/ray_current_cluster"
--- a/src/constants.jl
+++ b/src/constants.jl
@@ -37,7 +37,6 @@ const _check_msg = "Check that its definition is not implicitly capturing a larg
                    "array or other object in scope. Tip: use `Ray.put()` to put large " *
                    "objects in the Ray object store."
 
-
 # env var to control whether logs are sent do stderr or to file.  if "1", sent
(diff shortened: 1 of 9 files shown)


src/runtime.jl
```diff
@@ -355,7 +303,7 @@ julia -e 'using Ray; start_worker()' -- \
   --ray_redis_password= \
   --ray_session_dir=/tmp/ray/session_2023-08-09_14-14-28_230005_27400 \
   --ray_logs_dir=/tmp/ray/session_2023-08-09_14-14-28_230005_27400/logs \
-  --ray_node_ip_address=127.0.0.1
+  --ray_node_ip_ADDRESS=127.0.0.1
```
omus, line 306: Unintentional?

src/runtime.jl
```diff
@@ -125,79 +135,17 @@ Get the current task ID for this worker in hex format.
...
-    =#
-    line = open(joinpath(session_dir, "logs", "raylet.out")) do io
-        while !eof(io)
-            line = readline(io)
-            if contains(line, "Starting agent process")
-                return line
-            end
-        end
-    end
+function get_node_to_connect_for_driver()
```
omus, line 138: Tests? I would definitely pass in `GLOBAL_STATE_ACCESSOR[]` and `NODE_IP_ADDRESS`.

src/runtime.jl
```diff
@@ -99,7 +94,22 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
     job_config = JobConfig(RuntimeEnvInfo(runtime_env), metadata)
     serialized_job_config = _serialize(job_config)
 
-    ray_jll.initialize_driver(args..., job_id, logs_dir, serialized_job_config)
+    raylet, store, node_port = get_node_to_connect_for_driver()
+
+    @info begin
```
omus, line 99: Lost this comment in the shuffle
```suggestion
    # TODO: downgrade to debug
    # https://github.com/beacon-biosignals/Ray.jl/issues/53
    @info begin
```

omus (commented):
Nice, this is a good improvement over the orignal

test/global_state_accessor.jl
```diff
@@ -0,0 +1,24 @@
...
+
+    raylet_regex = r"^/tmp/ray/session_[0-9,_,-]+/sockets/raylet"
+    @test !isnothing(match(raylet_regex, String(raylet)))
+
+    store_regex = r"^/tmp/ray/session_[0-9,_,-]+/sockets/plasma_store"
+    @test !isnothing(match(store_regex, String(store)))
+
+    @test node_port isa Number
+
+    ray_jll.Disconnect(GLOBAL_STATE_ACCESSOR[])
```
glennmoy, line 23: I'll note tests still pass without explicitly `Disconnect`ing but I figured it's good practice anyway in case these tests are run in isolation

src/runtime.jl
```diff
@@ -1,5 +1,23 @@
...
+This is set during [`Ray.init`](@ref) and used there to get the the raylet name, object
+store name, node manager port, and the next job IDJob ID for the driver.
+"""
+const GLOBAL_STATE_ACCESSOR = Ref{ray_jll.GlobalStateAccessor}()
+
+# env var to control whether logs are sent do stderr or to file.  if "1", sent
+# to stderr; otherwise, will be sent to files in `/tmp/ray/session_latest/logs/`
+# https://github.com/beacon-biosignals/ray/blob/4ceb62daaad05124713ff9d94ffbdad35ee19f86/python/ray/_private/ray_constants.py#L198
+const LOGGING_REDIRECT_STDERR_ENVIRONMENT_VARIABLE = "RAY_LOG_TO_STDERR"
+
```
omus, line 14: Consider moving the non-`Ref` constants to a `constants.jl`

src/runtime.jl
```diff
@@ -67,16 +70,8 @@ function init(runtime_env::Union{RuntimeEnv,Nothing}=nothing;
...
-    # information instead of parsing logs?  I can't quite tell where it's coming
-    # from (set from a `ray.address` config option):
-    # https://github.com/beacon-biosignals/ray/blob/7ad1f47a9c849abf00ca3e8afc7c3c6ee54cda43/java/runtime/src/main/java/io/ray/runtime/config/RayConfig.java#L165-L171
-
-    # we use session_dir here instead of logs_dir since logs_dir can be set to
-    # "" to disable file logging without using env var
-    args = parse_ray_args_from_raylet_out(session_dir)
-    gcs_address = args[3]
+    # gcs-address=127.0.0.1:6379
+    gcs_address = read("/tmp/ray/ray_current_cluster", String)
```
omus, line 74:
```suggestion
    gcs_address = read("/tmp/ray/ray_current_cluster", String)  # host:port (e.g. 127.0.0.1:6379)
```

test/global_state_accessor.jl
```diff
@@ -0,0 +1,24 @@
...
+    # RAY_LOG(INFO) message about not finding a local Raylet with that address
+    # https://github.com/beacon-biosignals/ray/blob/448a83caf44108fc1bc44fa7c6c358cffcfcb0d7/src/ray/gcs/gcs_client/global_state_accessor.cc#L356-L359
+    raylet, store, node_port = get_node_to_connect_for_driver(GLOBAL_STATE_ACCESSOR[],
+                                                              NODE_IP_ADDRESS)
+
+    raylet_regex = r"^/tmp/ray/session_[0-9,_,-]+/sockets/raylet"
+    @test !isnothing(match(raylet_regex, String(raylet)))
+
+    store_regex = r"^/tmp/ray/session_[0-9,_,-]+/sockets/plasma_store"
+    @test !isnothing(match(store_regex, String(store)))
```
omus, line 19: Regex character groups don't use a comma separator:
```julia
julia> occursin(r"^[0-9_-]+$", "0123456789_-")
true

julia> occursin(r"^[0-9_-]+$", "0123456789_-,")
false

julia> occursin(r"^[0-9,_,-]+$", "0123456789_-,")
true
```
```suggestion
    raylet_regex = r"^/tmp/ray/session_[0-9_-]+/sockets/raylet"
    @test occursin(raylet_regex, String(raylet))

    store_regex = r"^/tmp/ray/session_[0-9,_,-]+/sockets/plasma_store"
    @test occursin(store_regex, String(store))
```

test/global_state_accessor.jl
```diff
@@ -0,0 +1,24 @@
+@testset "GlobalStateAccessor" begin
+    using Ray: GLOBAL_STATE_ACCESSOR, NODE_IP_ADDRESS, get_node_to_connect_for_driver
+
+    gcs_address = read("/tmp/ray/ray_current_cluster", String)
```
omus, line 4: May be worth making this a constant

omus (approved):
Looks great! Couple of suggestions

test/global_state_accessor.jl
glennmoy: I was gonna do the same with the `session_dir` as well

src/constants.jl
```diff
@@ -0,0 +1,51 @@
...
+const FUNCTION_SIZE_ERROR_THRESHOLD = 100_000_000  # in bytes
+
+# python uses "fun" for the namespace: https://github.com/beacon-biosignals/ray/blob/7ad1f47a9c849abf00ca3e8afc7c3c6ee54cda43/python/ray/_private/ray_constants.py#L380
+# so "jlfun" seems reasonable
+const FUNCTION_MANAGER_NAMESPACE = "jlfun"
+
+# https://github.com/beacon-biosignals/ray/blob/1c0cddc478fa33d4c244d3c30aba861a77b0def9/python/ray/_private/utils.py#L744-L746
+const _check_msg = "Check that its definition is not implicitly capturing a large " *
+                   "array or other object in scope. Tip: use `Ray.put()` to put large " *
+                   "objects in the Ray object store."
```
omus, line 38: I'd keep this where it was as it's only used by one function

test/global_state_accessor.jl
```diff
@@ -0,0 +1,25 @@
...
+    # Note: passing in a fake node IP address won't throw an error, instead it reports a
+    # RAY_LOG(INFO) message about not finding a local Raylet with that address
+    # https://github.com/beacon-biosignals/ray/blob/448a83caf44108fc1bc44fa7c6c358cffcfcb0d7/src/ray/gcs/gcs_client/global_state_accessor.cc#L356-L359
+    raylet, store, node_port = get_node_to_connect_for_driver(GLOBAL_STATE_ACCESSOR[],
+                                                              NODE_IP_ADDRESS)
+
+    raylet_regex = r"^/tmp/ray/session_[0-9_-]+/sockets/raylet"
+    @test occursin(raylet_regex, String(raylet))
+
+    store_regex = r"^/tmp/ray/session_[0-9,_,-]+/sockets/plasma_store"
```
omus, line 19: Whoops! Did I make a bad suggestion?
```suggestion
    store_regex = r"^/tmp/ray/session_[0-9_-]+/sockets/plasma_store"
```

omus (approved):

test/global_state_accessor.jl
glennmoy: I didn't catch it either :D

omus:
> The changes I just introduced I should just assist in debugging the underlying failure.

Only information gathered from these failures is that the `process_eval` failure I noticed doesn't happen consistently. Updating this PR to include #223 which hopefully provides some new information.

omus (approved):

omus:
I guess we'll have to just proceed. My latest theory is that the `StdString`s returned in `get_node_to_connect_for_driver` are sometimes being deallocated before they are used by the C++ core worker. If that is correct the the two full passes we've seen from #223 have been a coincidence.

omus:
After running some additional experiments in #224 I'm am confident the issue seen in this PR was related to a stale GHA cache. When this PR failed it failed consistently for 2/4 CI jobs and after this PR was updated with `main` the PR started reliably passing. It's unfortunate we never saw the failure with #223 as it could have been interesting to get more information but this appears to have been caused by updating CxxWrap in #217. I ran into some issues locally with this but usually the failure occurred immediately upon loading the wrong library. 

At one point I did wipe out the GHA cache for this PR and I thought the issue returned but possibly I made a mistake there or the cache from `main` was also bad.

In any case if I'm wrong about this and there's a lurking bug we have better reporting in place to investigate further.

glennmoy:
> After running some additional experiments in https://github.com/beacon-biosignals/Ray.jl/pull/224 I'm am confident the issue seen in this PR was related to a stale GHA cache.

Interesting... I thought I deleted the cache as well and when I re-ran there was an initial cache miss but the test still failed so it must have restored from a corrupted backup

miss:
https://github.com/beacon-biosignals/Ray.jl/actions/runs/6641903107/job/18045441825#step:7:64

backup:
https://github.com/beacon-biosignals/Ray.jl/actions/runs/6641903107/job/18045441825#step:8:1135

src/constants.jl
```diff
@@ -0,0 +1,46 @@
...
+# ...but that it's not stable for some reason.  but.....neither is a random
+# UUID?????
+#
+# the function table key is built like
+# <key type>:<jobid>:key
+
+# function manager holds:
+# local cache of functions (keyed by function id/hash from descriptor)
+# gcs client
+# ~~maybe job id?~~ this is managed by the core worker process
```
kleinschmidt, line 25: these are pretty specific to the function manager itself, not the constants that got moved here.

src/runtime.jl
```diff
@@ -125,79 +122,17 @@ Get the current task ID for this worker in hex format.
...
-    store_match = match(r"object-store-name=((\/[a-z,0-9,_,-]+)+)", line)
-    store = if store_match !== nothing
-        String(store_match[1])
-    else
-        error("Unable to find Object Store socket")
-    end
+function get_node_to_connect_for_driver(global_state_accessor, node_ip_address)
+    node_to_connect = StdString()
+    status = ray_jll.GetNodeToConnectForDriver(global_state_accessor, node_ip_address,
+                                               CxxPtr(node_to_connect))
```
kleinschmidt, line 128: it's probably a good idea to check here whether the status is good; if the GCS can't be reached after a timeout (something like 60s), it'll terminate the whole process, and that failure can't be detected when calling `Connect` (it tries to connect in a separate thread)

kleinschmidt: I guess you'd fail to like parse/destructure the node info though if it the call returned junk

PR #215, opened 2023-10-20
pass `max_retries` during task submission

### LINKED ISSUE #143: Improve handling of tasks killed due to low memory
While running some multi-node benchmarks I noticed this failure occur at the end of the run on the driver:

```
ERROR: LoadError: TaskFailedException

    nested task error: ArgumentError: 'Int32' iterates 'Int32' values, which doesn't satisfy the Tables.jl `AbstractRow` interface
    Stacktrace:
     [1] invalidtable(#unused#::Int32, #unused#::Int32)
       @ Tables ~/.julia/packages/Tables/AcRIE/src/tofromdatavalues.jl:41
     [2] iterate
       @ ~/.julia/packages/Tables/AcRIE/src/tofromdatavalues.jl:47 [inlined]
     [3] buildcolumns
       @ ~/.julia/packages/Tables/AcRIE/src/fallbacks.jl:209 [inlined]
     [4] _columns
       @ ~/.julia/packages/Tables/AcRIE/src/fallbacks.jl:274 [inlined]
     [5] columns
       @ ~/.julia/packages/Tables/AcRIE/src/fallbacks.jl:258 [inlined]
     [6] DataFrames.DataFrame(x::Int32; copycols::Bool)
       @ DataFrames ~/.julia/packages/DataFrames/LteEl/src/other/tables.jl:57
     [7] append!(df::DataFrames.DataFrame, table::Int32; cols::Symbol, promote::Bool)
       @ DataFrames ~/.julia/packages/DataFrames/LteEl/src/other/tables.jl:71
    ...

...and 6 more exceptions.

Stacktrace:
 [1] sync_end(c::Channel{Any})
   @ Base ./task.jl:445
 [2] macro expansion
   @ ./task.jl:477 [inlined]
 [3] reduce_results(ray_objects::Vector{Any})
    ...
in expression starting at /tmp/ray/session_2023-09-22_07-13-21_412745_8/runtime_resources/working_dir_files/_ray_pkg_0a43fd8298c1f456/migration.jl:5
```
Looking at the Ray dashboard for this job I saw 7 failed tasks with this error:
```
Error Type: OUT_OF_MEMORY
Task was killed due to the node running low on memory.
Memory on the node (IP: 10.0.18.21, ID: c1e1fd57171f3c092f46167a78eb2d5007398539847a38edfe7f1829) where the task (task ID: 98937e2987c30bec7128cff4045939da2639045b1f000000, name=v0_4_4.process_segment, pid=162, memory used=3.10GB) was running was 51.31GB / 54.00GB (0.950157), which exceeds the memory usage threshold of 0.95. Ray killed this worker (ID: a4340f66d387df4707564760c0913885d662d19424f2ba87c991daf3) because it was the most recently scheduled task; to see more information about memory usage on this node, use `ray logs raylet.out -ip 10.0.18.21`. To see the logs of the worker, use `ray logs worker-a4340f66d387df4707564760c0913885d662d19424f2ba87c991daf3*out -ip 10.0.18.21. Top 10 memory users:
PID	MEM(GB)	COMMAND
161	4.24	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
159	3.53	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
163	3.52	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
432	3.42	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
423	3.33	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
402	3.31	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
164	3.26	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
503	3.25	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
158	3.13	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
162	3.10	/usr/local/julia/bin/julia -Cnative -J/usr/local/julia/lib/julia/sys.so -g1 -e using Ray; start_work...
Refer to the documentation on how to address the out of memory issue: https://docs.ray.io/en/latest/ray-core/scheduling/ray-oom-prevention.html. Consider provisioning more memory on this node or reducing task parallelism by requesting more CPUs per task. Set max_retries to enable retry when the task crashes due to OOM. To adjust the kill threshold, set the environment variable `RAY_memory_usage_threshold` when starting Ray. To disable worker killing, set the environment variable `RAY_memory_monitor_refresh_ms` to zero.
```
It would be better if more details in this case were reported back to the driver. Additionally, I'm not sure why these tasks were not rescheduled.

### LINKED ISSUE #213: Allow users to set `max_retries` when submitting a task
We currently hardcode this as 0 here: 

https://github.com/beacon-biosignals/Ray.jl/blob/836b43b580faae468ffbd15ac66cb54651e1a714/build/wrapper.cc#L174

I think it's a simple matter of passing through an Int32 in `submit_task`.

From digging into the actual ray source a bit, I think the other apparently relevant task option (`retry_exceptions` and `serialized_retry_exception_allowlist`) are for retrying _application_ errors, which we don't want to handle now.  I'm not 100% sure about the `retry_exceptions` bit but from a bit of digging, it seems like it's only checked when the allowlist is consulted in the raylet code...

https://github.com/beacon-biosignals/ray/blob/015518473a40997c1ee1591c24c65483377971cc/src/ray/core_worker/core_worker.h#L834-L837

kleinschmidt:
Fixes #213 and may fix #143 

Given that we're still forbidding retries of application errors, I'm not quite sure how to test this.

### DIFF
--- a/build/wrapper.cc
+++ b/build/wrapper.cc
@@ -144,7 +144,8 @@ std::vector<std::shared_ptr<RayObject>> cast_to_task_args(void *ptr) {
 ObjectID _submit_task(const ray::JuliaFunctionDescriptor &jl_func_descriptor,
                       const std::vector<TaskArg *> &task_args,
                       const std::string &serialized_runtime_env_info,
-                      const std::unordered_map<std::string, double> &resources) {
+                      const std::unordered_map<std::string, double> &resources,
+                      int max_retries) {
 
     auto &worker = CoreWorkerProcess::GetCoreWorker();
 
@@ -171,8 +172,8 @@ ObjectID _submit_task(const ray::JuliaFunctionDescriptor &jl_func_descriptor,
         func,
         args,
         options,
-        /*max_retries=*/0,
-        /*retry_exceptions=*/false,
+        max_retries,                // application error counter, gate for OOM retry (!=0)
+        /*retry_exceptions=*/false, // only applies to application errors
         scheduling_strategy,
         /*debugger_breakpoint=*/""
     );
--- a/src/ray_julia_jll/common.jl
+++ b/src/ray_julia_jll/common.jl
@@ -373,10 +373,11 @@ end
 # XXX: Need to convert julia vectors to StdVector and build the
 # `std::unordered_map` for resources. This function helps us avoid having
 # CxxWrap as a direct dependency in Ray.jl
-function _submit_task(fd, args, serialized_runtime_env_info, resources::AbstractDict)
+function _submit_task(fd, args, serialized_runtime_env_info, resources::AbstractDict,
+                      max_retries)
     @debug "task resources: " resources
     resources = build_resource_requests(resources)
-    return _submit_task(fd, args, serialized_runtime_env_info, resources)
+    return _submit_task(fd, args, serialized_runtime_env_info, resources, max_retries)
 end
 
 # work around lack of wrapped `std::unordered_map`
--- a/src/runtime.jl
+++ b/src/runtime.jl
@@ -140,7 +140,8 @@ initialize_coreworker_driver(args...) = ray_jll.initialize_coreworker_driver(arg
 # TODO: Move task related code into a "task.jl" file
 function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
                      runtime_env::Union{RuntimeEnv,Nothing}=nothing,
-                     resources::Dict{String,Float64}=Dict("CPU" => 1.0))
+                     resources::Dict{String,Float64}=Dict("CPU" => 1.0),
+                     max_retries::Integer=0)
     export_function!(FUNCTION_MANAGER[], f, get_job_id())
     fd = ray_jll.function_descriptor(f)
     task_args = serialize_args(flatten_args(args, kwargs))
@@ -155,7 +156,8 @@ function submit_task(f::Function, args::Tuple, kwargs::NamedTuple=NamedTuple();
         ray_jll._submit_task(fd,
                              transform_task_args(task_args),
                              serialized_runtime_env_info,
-                             resources)
+                             resources,
+                             max_retries)
     end
     # CoreWorker::SubmitTask calls TaskManager::AddPendingTask which initializes
     # the local ref count to 1, so we don't need to do that here.


kleinschmidt:
This is pretty simple but is going to be hard to test, so I'm going to open it for review despite lacking functional tests that tasks are _actually_ retried as expected.

glennmoy (approved):
Changes LGTM

glennmoy:
I wonder for the sake of testing do we want to allow and then force an applicatoin error? or do we want to be very strict about guarding against that?

kleinschmidt:
> force an applicatoin error

those aren't retried at all right now, only OOMs.  I played around a bit with trying to force an OOM but that's pretty fragile...

kleinschmidt:
> > > force an applicatoin error
> > 
> > 
> > those aren't retried at all right now, only OOMs. I played around a bit with trying to force an OOM but that's pretty fragile...
> 
> what if you used mocking to`throw(OutOfMemoryError)`?

Alas, that would be picked up as an application error.  Ray has its own monitor that detects memory pressure on the whole node and starts killing tasks when a certain threshold is reached.  So you really do need to trigger a (near) OOM to activate it.

The motivation for this is that the OS doesn't care which process is which and will kill whatever it wants.  So it's better for Ray to preemptively kill things that are known to replaceable (i.e., workers running tasks) to avoid triggering the OS OOM killer if at all possible: https://docs.ray.io/en/latest/ray-core/scheduling/ray-oom-prevention.html

glennmoy:
> > > > force an applicatoin error
> > > 
> > > 
> > > those aren't retried at all right now, only OOMs. I played around a bit with trying to force an OOM but that's pretty fragile...
> > 
> > 
> > what if you used mocking to`throw(OutOfMemoryError)`?
> 
> Alas, that would be picked up as an application error. Ray has its own monitor that detects memory pressure on the whole node and starts killing tasks when a certain threshold is reached. So you really do need to trigger a (near) OOM to activate it.
> 
> The motivation for this is that the OS doesn't care which process is which and will kill whatever it wants. So it's better for Ray to preemptively kill things that are known to replaceable (i.e., workers running tasks) to avoid triggering the OS OOM killer if at all possible: https://docs.ray.io/en/latest/ray-core/scheduling/ray-oom-prevention.html

ah I see! oh well what can we do

PR #217, opened 2023-10-23
Update to use CxxWrap 0.14.1

omus:
This release includes the `StdString` changes from https://github.com/JuliaInterop/CxxWrap.jl/pull/381 and https://github.com/JuliaInterop/CxxWrap.jl/pull/378 which allows us to drop the `safe_convert` function which was used to work around the issues noticed with binary string data handling (introduced in https://github.com/beacon-biosignals/Ray.jl/pull/203).

Overall this leads to cleaner code and potentially could fix any lurking issues with `StdString` binary data we weren't already handling.

Developers running local setups should run this after updating:
```
julia --project=build -e 'using Pkg; Pkg.update()' && julia --project=build build/build_library.jl && julia --project -e 'using Pkg; Pkg.update(); using Ray'
```


Project.toml
```diff
@@ -20,11 +20,11 @@ libcxxwrap_julia_jll = "3eaa8342-bff7-56a5-9981-c04077f7cee7"
 [compat]
 Aqua = "0.5"
 ArgParse = "1"
-CxxWrap = "0.13.4"
+CxxWrap = "0.14.1"
 JSON3 = "1"
 LoggingExtras = "1"
 julia = "1.8"
-libcxxwrap_julia_jll = "0.9.7"
+libcxxwrap_julia_jll = "0.11"
```
glennmoy, line 27: is this a driveby bump?

src/ray_julia_jll/common.jl
```diff
@@ -157,21 +157,17 @@ let msg_types = (Address, JobConfig, ObjectReference)
...
 end
 
 function Serialization.serialize(s::AbstractSerializer, message::Message)
-    serialized_message = safe_convert(String, SerializeAsString(message))
+    serialized_message = SerializeAsString(message)
 
     serialize_type(s, Message)
     serialize(s, supertype(typeof(message)))
-    serialize(s, serialized_message)
+    serialize(s, String(serialized_message))
```
glennmoy, line 170: do we need the `String` here?

src/ray_julia_jll/ray_julia_jll.jl
```diff
@@ -11,13 +11,12 @@ using libcxxwrap_julia_jll
 
 abstract type BaseID end
 
-@wrapmodule(joinpath(artifact"ray_julia", "julia_core_worker_lib.so"))
+@wrapmodule(() -> joinpath(artifact"ray_julia", "julia_core_worker_lib.so"))
```
glennmoy, line 14: what's this adding?

glennmoy (approved):

Project.toml
omus: So CxxWrap requires this minimum version so I also bumped this

src/ray_julia_jll/common.jl
omus: Yes, serializing a `StdString` doesn't work at the moment. See: https://github.com/JuliaInterop/CxxWrap.jl/pull/381#issuecomment-1773402995

src/ray_julia_jll/ray_julia_jll.jl
omus: This is required by newer versions of CxxWrap:
```
┌ Error: Calling `@wrapmodule` with the path of the library to load is no longer supported.
│ Pass the name of a function returning the path instead, e.g. use `libfoo_jll.get_libfoo_path` instead of `libfoo_jll.libfoo`.
└ @ CxxWrap.CxxWrapCore ~/.julia/packages/CxxWrap/66SRr/src/CxxWrap.jl:729
```
I didn't look into this but it looks to me like its for better support for BinaryBuilder libraries

PR #220, opened 2023-10-24
Release process follow ups

### LINKED ISSUE #210: Pre-Release.yml follow ups
- [ ] #206 
- [ ] #208 
- [ ] Set as “latest release” in `Publish Release`
- [ ] `workflow_dispatch` should take as an input the commit sha to use as `tag`  (to enable backports)
- [ ] Pre-Release workflow should check project.version !in tags to allow back ports
- [ ] Investigate why the `artifact_CI.yml` workflow isn't being triggerd by `workflow_dispath` 
   - Perhaps we need to use BeaconBuddy token in tag step?

glennmoy:
This knocks out some of items in #210 namely:

- workflow_dispatch should take as an input the commit sha to use as tag (to enable backports)
- Pre-Release workflow should check project.version !in tags to allow back ports
- Set as “latest release” in Publish Release
- release are based on changes since previous tag 
- stable documenter builds are based on Releases

Not solved yet:
- how to automatically trigger `artifacts_CI` workflow after triggering `Pre_release CI` using `workflow_dispatch` (seems to ignore that it's a workflow_dispatch and doesn't trigger it)

---
  - https://app.asana.com/0/0/1205785985070962

### DIFF
--- a/.github/workflows/Pre_release.yml
+++ b/.github/workflows/Pre_release.yml
@@ -88,10 +88,11 @@ jobs:
           git push origin $tag
       - name: Publish Pre-Release
         uses: ncipollo/release-action@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.DOCUMENTER_KEY }}
         with:
           prerelease: true
           tag: ${{ steps.project_toml.outputs.tag }}
           commit: ${{ steps.commit.outputs.sha }}
           artifacts: ${{ env.tarballs_dir }}/*.tar.gz
-          token: ${{ secrets.DOCUMENTER_KEY }}
           generateReleaseNotes: true
--- a/.github/workflows/Pre_release.yml
+++ b/.github/workflows/Pre_release.yml
@@ -66,10 +66,6 @@ jobs:
         working-directory: ${{ env.build_dir }}
       - name: Commit Changes and Push Tag
         id: commit
-        env:
-          # Avoid using the workflows GITHUB_TOKEN as it will not re-trigger CI jobs
-          # https://github.community/t/push-from-action-does-not-trigger-subsequent-action/16854
-          GITHUB_TOKEN: ${{ secrets.DOCUMENTER_KEY }}
         run: |
           # Alternatively environment variables can be used but both author/committer need to be set
           # https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing
--- a/.github/workflows/Pre_release.yml
+++ b/.github/workflows/Pre_release.yml
@@ -31,8 +31,6 @@ jobs:
       - name: Check for new Project version
         id: project_toml
         shell: julia --color=yes --project {0}
-        env:
-          LATEST_TAG: ${{ steps.latest_tag.outputs.tag }}
         run: |
           using Pkg.Types
           project = read_project("Project.toml")
--- a/.github/workflows/Pre_release.yml
+++ b/.github/workflows/Pre_release.yml
@@ -86,11 +86,11 @@ jobs:
           git tag $tag $(git rev-parse HEAD)
           git push origin $tag
       - name: Publish Pre-Release
-        uses: softprops/action-gh-release@v1
+        uses: ncipollo/release-action
         with:
           prerelease: true
-          tag_name: ${{ steps.project_toml.outputs.tag }}
-          target_commitish: ${{ steps.commit.outputs.sha }}
-          generate_release_notes: true
-          files: ${{ env.tarballs_dir }}/*.tar.gz
+          tag: ${{ steps.project_toml.outputs.tag }}
+          commit: ${{ steps.commit.outputs.sha }}
+          artifacts: ${{ env.tarballs_dir }}/*.tar.gz
           token: ${{ secrets.BEACON_BUDDY_GH_TOKEN }}
+          generateReleaseNotes: true
--- a/.github/workflows/Pre_release.yml
+++ b/.github/workflows/Pre_release.yml
@@ -22,15 +22,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 0  # Needed determine latest tag
+          fetch-depth: 0  # Needed determine tags
           persist-credentials: true
           ref: ${{ inputs.commit }}
-      - name: Get Latest Tag
-        id: latest_tag
-        uses: WyriHaximus/github-action-get-previous-tag@v1
-        # TODO: in future we might need to perform backports, at which point we'll need to check
-        # that the project.version has not already been tagged 
-      - name: Check for updated Project version
+      - name: Check for new Project version
         id: project_toml
         shell: julia --color=yes --project {0}
         env:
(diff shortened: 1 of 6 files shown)


.github/workflows/Pre_release.yml
```diff
@@ -1,6 +1,14 @@
 name: "Publish Pre-Release"
 on:
   workflow_dispatch:
+    inputs:
+      commit:  
+        # Note: The tag itself will not be based off this commit. As we must first commit the 
+        # updates to the Artifacts.toml, then the tag will be based off that commit.
+        description: 'The commit to branch from when making the tag'
```
omus, line 8: YAML requires very little quoting. If you do want to quote this stick to double-quotes. Single quotes are only required for GHA conditionals
```suggestion
        description: The commit to branch from when making the tag
```

.github/workflows/Pre_release.yml
```diff
@@ -12,29 +20,25 @@ jobs:
...
-        uses: WyriHaximus/github-action-get-previous-tag@v1
-        # TODO: in future we might need to perform backports, at which point we'll need to check
-        # that the project.version has not already been tagged 
-      - name: Check for updated Project version
+          ref: ${{ inputs.commit }}
+      - name: Check for new Project version
         id: project_toml
         shell: julia --color=yes --project {0}
         env:
           LATEST_TAG: ${{ steps.latest_tag.outputs.tag }}
```
omus, line 35:
```suggestion
```

.github/workflows/Pre_release.yml
```diff
@@ -12,29 +20,25 @@ jobs:
...
         run: |
           using Pkg.Types
           project = read_project("Project.toml")
-          latest_tag = parse(VersionNumber, ENV["LATEST_TAG"])
+          tags = parse.(VersionNumber, split(readchomp(`git tag -l`), "\n"))
```
omus, line 39: Just curious if you verified this? I thought maybe we'd have to do `git fetch --tags`

.github/workflows/Pre_release.yml
```diff
@@ -12,29 +20,25 @@ jobs:
...
           @info "Project version: project.version"
-          @info "Latest tag: latest_tag"
-          project.version > latest_tag || exit(1)  # error if can't make a tag
+          project.version in tags && error("Tag v$(project.version) already exists")
```
omus, line 41: Doesn't `project.version` lack the `v` suffix while `tags` would include it?

.github/workflows/Pre_release.yml
```diff
@@ -12,29 +20,25 @@ jobs:
...
     name: "Publish Pre-Release"
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 0  # Needed determine latest tag
+          fetch-depth: 0  # Needed determine tags
           persist-credentials: true
```
omus, line 29: Could be part of your problem:
```suggestion
          persist-credentials: false  # Avoid using these credentials as pushing new commits doesn't re-trigger CI jobs
```

.github/workflows/Pre_release.yml
glennmoy: yup it seems to work https://github.com/beacon-biosignals/Ray.jl/actions/runs/6630217614/job/18011182410?pr=220#step:3:18

.github/workflows/Pre_release.yml
glennmoy: no it get's parsed as a `VersionNumber` 
```julia
julia> project.version
v"0.0.4-dev"
```

glennmoy:
looking into using `worflow_run` as suggested elsewhere, I don't think this will work as it only runs the workflow against branches and not tags https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches

.github/workflows/Pre_release.yml
omus: TIL:
```julia
julia> parse(VersionNumber, "v0.0.1")
v"0.0.1"

julia> parse(VersionNumber, "0.0.1")
v"0.0.1"
```

omus:
> looking into using `worflow_run` as suggested elsewhere, I don't think this will work as it only runs the workflow against branches and not tags https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#limiting-your-workflow-to-run-based-on-branches

Do we need to use `workflow_run` on tags? Why not use `push.tags` as the event?

glennmoy:
> Do we need to use `workflow_run` on tags? Why not use `push.tags` as the event?

We don't need to, I was just exploring it as `push.tags` isn't being triggered for some reason, even though when triggered via `workflow_dispatch` it should allow downstream workflows to be triggered https://github.blog/changelog/2022-09-08-github-actions-use-github_token-with-workflow_dispatch-and-repository_dispatch/

.github/workflows/Pre_release.yml
```diff
@@ -1,6 +1,14 @@
 name: "Publish Pre-Release"
 on:
   workflow_dispatch:
+    inputs:
+      commit:  
+        # Note: The tag itself will not be based off this commit. As we must first commit the 
+        # updates to the Artifacts.toml, then the tag will be based off that commit.
+        description: The commit to branch from when making the tag
+        required: true
+  push:
```
glennmoy, line 10:
```suggestion
```
used for testing

.github/workflows/Pre_release.yml
```diff
@@ -80,10 +82,9 @@ jobs:
           git tag $tag ${{ steps.commit.outputs.sha }}
           git push origin $tag
       - name: Publish Pre-Release
-        uses: softprops/action-gh-release@v1
+        uses: ncipollo/release-action@v1
```
glennmoy, line 85: switching to this action as it allows us to set the rlease as "latest release"

glennmoy:
I think the reason the release notes are whack is because our tag is not targetting the `main` branch

https://github.com/orgs/community/discussions/5975#discussioncomment-1799743

>One of the issues folks have been running into here are repositories that are using advanced release strategies with release branches. The new regex is not going to help in those cases.
>Our release generating tool is built on top of Pull-Requests and requires that your PRs target the same target branch that the release is created from.

glennmoy:
I'll note I'm following along with https://github.com/ncipollo/release-action/issues/198 to see if there's a workaround

glennmoy:
I've found https://github.com/mikepenz/release-changelog-builder-action and will look to generate the changelog from that

glennmoy:
Ok I'm calling this a win

Screenshot 2023-11-01 at 17 53 01

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Ray"
 uuid = "3f779ece-f0b6-4c4f-a81a-0cb2add9eb95"
 authors = ["Beacon Biosignals, Inc"]
-version = "0.0.3"
+version = "0.0.4-dev"
```
glennmoy, line 4:
```suggestion
version = "0.0.3"
```

.github/workflows/Pre_release.yml
omus: If we create a tag that isn't a version number this will break. We should do:
```suggestion
          tags = split(readchomp(`git tag -l`), "\n")
```
and do a string comparison later

.github/workflows/Pre_release.yml
```diff
@@ -10,31 +17,25 @@ env:
...
-          LATEST_TAG: ${{ steps.latest_tag.outputs.tag }}
         run: |
           using Pkg.Types
           project = read_project("Project.toml")
-          latest_tag = parse(VersionNumber, ENV["LATEST_TAG"])
+          tags = parse.(VersionNumber, split(readchomp(`git tag -l`), "\n"))
           @info "Project version: project.version"
-          @info "Latest tag: latest_tag"
-          project.version > latest_tag || exit(1)  # error if can't make a tag
+          project.version in tags && error("Tag v$(project.version) already exists")
```
omus, line 38:
```suggestion
          project_version_tag = "v$(project.version)"
          project_version_tag in tags && error("Tag $project_version_tag already exists")
```

omus (approved):
One last thing

PR #247, opened 2024-07-05
Update ray coreworker library to 2.31.0

omus:


omus:
Retrieving the core worker address/port has changed and we'll need to update that code which was fragile to begin with.

omus:
I'm now seeing some issues just with running `ray start --head` so maybe something else is going on too. I'll have to pick this up later.

omus:
Will need to make pre-built Ray CLI binaries. We have some abandoned instructions here: https://github.com/beacon-biosignals/ray/pull/11

